Issue #170, opened 2015-04-25
Undoing Expense causes Exception


cjrt:
If you undo an expense, but have already removed the item it will throw an exception, as the object it was pointing to no longer exists.

Object reference not set to instance of object.

System.NullRefference.Exception: Object reference not set to instance of object
  at Chummer.frmCareer.tsUndoNuyenExpense_Click(Object sender, EventArgs e) in m:\Coding\chummer5a\Chummer\frmCareer.cs:line 13820

If I knew enough C# to trivially add check for null I would make a pull request, but I have already wasted today playing Shadowrun when I have exams coming up.

Plus, if the object has already been removed should they be able to undo the expense? Or can we assume that if they're trying to refund the expense they have fair reason? Perhaps a pop up to say the item has already been removed, do they want to just revert the expense? These are design decisions I am less confident in making.

chummer5a:
Removing an item is already supposed to remove the improvement. That's a
bit weird.

cjrt:
Oh is it? How do I remove something without refunding it then? i.e. I buy a
cyber deck, it gets trashed/I lose it, I don't want that to be refunded
when I remove it from inventory?

DeastinY:
As far as I understood that's the default behavior. You delete the items and don't get the money - you delete the expense and you get the money back. At least that's what I'd expect (and what I verified so far). When removing expenses you always refund, as this is what removing an expense does. And if you just delete your cyberdeck from the geartab it's that you simply don't have the thing anymore (you still bought it once, so the expense is still there)

chummer5a:
Yeah, I had a brainfart there, was thinking of improvements. I'm having trouble replicating the problem in the first place though. No particular reason not to merge in the null check as Chummer does tend to be rather fragile in that regard, but we could be suppressing a larger issue.

cjrt:
That's what I expected. I thought he was saying that deleting the item
should delete the expense.

DeastinY:
ah okay

chummer5a:
Unable to reproduce the issue in current version. Closed, pending something breaking.

PR #172, opened 2015-04-26
Fixed Issue #170, Added German translations

DeastinY:
Sorry for whitespace issues. Did not find any solution to it yet (using VS2013 Community)


chummer5a:
All good; in the future, if you're making changes to a data file, can you increment the version number by one? ie -497 is current, increment to -496. It's needed for automatic updates; needs to be done in the manifestdata.xml and manifestlang.xml files as well. At some point I'll probably find a way to automatically increment it, but for now it's a manual process.

DeastinY:
Thanks for the info, will remember !
(Glad to help, thanks for doing this all in the first place !)

PR #242, opened 2015-06-18
Fix so that automatic mandatory qualities do not count towards Karma …

richard-zang:
…limit on chargen.

I added one line <code>objQuality.ContributeToLimit = false;</code> for every location where qualities were being automatically added as a result of a particular metatype/metavariant choice.

### DIFF
--- a/Chummer/Character Creation/frmKarmaMetatype.cs
+++ b/Chummer/Character Creation/frmKarmaMetatype.cs
@@ -633,6 +633,7 @@ namespace Chummer
 						if (objXmlQualityItem.Attributes["removable"] != null)
 							objSource = QualitySource.MetatypeRemovable;
 						objQuality.Create(objXmlQuality, _objCharacter, objSource, objNode, objWeapons, objWeaponNodes, strForceValue);
+                        objQuality.ContributeToLimit = false;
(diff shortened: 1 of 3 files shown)


richard-zang:
That's weird. It says I made a whole bunch of additions and a whole bunch of deletions when I all I did was add four lines of code to each file. Does it diff based on formatting? And would Visual Studio screw with formatting?

And how do I see the diff of the pull request?

Kromgar:
Wait... that wasn't a intended part of chummer? Holy shit

joha4270:
>And how do I see the diff of the pull request?

Just navigate to changed files. The reason it won't show is because the entire files have been replaced.

Speaking of that, visual studio uses spaces for formatting by default (and will offer to change the entire file to that on opening a file). To git, that is a new line.

I'm unsure how easy it is to go back a commit and reverse to non spaced file.

richard-zang:
Is the 25 karma towards positive qualities considered what you spent or what you have? Because from looking at /r/hubchargen racial traits that are qualities don't count those qualities as having been bought so they don't count towards the positive/negative limit.

richard-zang:
Okay I'm going to close this pull and try to resubmit a non-wonky formatted version.

PR #243, opened 2015-06-18
Karma limits for purchasing +/- qualities do not include racial traits

richard-zang:
I resubmitted this because the previous pull request had reformatted files which threw git off when diffing.

The reason for this pull request is that racial traits are not considered bought so therefore they should not be included in the 25 karma limit for purchasing positive and/or negative qualities.

Currently, racial traits for metatypes don't count towards the 25 karma limit because they are set to cost 0 when added to the character sheet.

Either the changes I've requested or by setting the karma cost to 0 for metagenetic would work. Although I think my way is better because it takes advantage of the ContributeToLimit that already exists.


joha4270:
Thanks for fixing your own bugs, gives me more time to work on new features. :D

Issue #318, opened 2015-08-02
Life Modules: spread ranks among several skills.


angelforest:
1st case:

> Academic Knowledge: [Ancient Language/Philosophy/Religion] +13 (spread out among at least three subjects, minimum three ranks per subject)

Okay, [this way](http://pastebin.com/PMCJZdGC) we can ensure that user have 3 ranks in 3 skills, but still can't check if he has 3 ranks in 4th skill, if he willing to take it instead of rising previous ones.
Also, it just looks ugly.

2nd case:

> Academic Knowledge: [Ancient Language/Philosophy/Religion] +13 (divide among all three, with no one skill having more than 7 ranks)

or even weirder variant in celebrity module:

> +6 to buy any skills you think appropriate to your field of celebrity (no more than 3 ranks in any
> individual skill; ranks in Con and Escape Artist may be purchased with these points)

In that case there is no minimum number of skills and ranks in them, just maximum. And we can't enforce maximum with only data entry.

Related stuff:

> Charisma +1, and +1 to two attributes of your choice (must be two separate attributes, other than Charisma)

Need somehow check that there two different attributes or at least put note about that.

joha4270:
>Academic Knowledge: [Ancient Language/Philosophy/Religion] +13 (spread out among at least three subjects, minimum three ranks per subject)

This can be enforced by providing only a <name> Ancient Language</name> instead of <option>

That will force that skill to be that specific powers, he can still hack his way around, but people would have to look for it. Don't belive it is worth doing something more advanced at present time as the time to implement it would be large.

>Academic Knowledge: [Ancient Language/Philosophy/Religion] +13 (divide among all three, with no one skill having more than 7 ranks)   

Afraid we can't enforce that. Not sure (need a look) how hard it would be to add a specific max, but that is probably being per skill, not from this specific one. (Is there a path where you can get it the skill from other places or is that without worry?)

>+6 to buy any skills you think appropriate to your field of celebrity (no more than 3 ranks in any
individual skill; ranks in Con and Escape Artist may be purchased with these points)

Afraid this one can't be implemented without a few thousand lines of new code, first iteration is probably going to look for honest players, or a check when marking a character as created. (A tag about a knoskill being from life module might also solve this with @2.

>Charisma +1, and +1 to two attributes of your choice (must be two separate attributes, other than Charisma)

[...]

PR #320, opened 2015-08-02
Life modules data

angelforest:
All modules from run faster, data trails and chrome flesh.
I noted problematic places, still not sure about solution in #318, so I didn't change this.


chummer5a:
Awesome, thanks for the quick work!

PR #355, opened 2015-08-26
Added AI qualities

SydneySedai:
Not familiar with the xml and backend stuff going on here, so a few things are missing from Bonus entries
Also, apologies in advance if I did this crap all wrong, because I'm just working by example from what I saw in existing XML.
Not a clue what the ID field is, so I left them blank.

- No idea how to get Designer to apply its bonuses to a specific device; probably best to just add a text field for the user to enter their home device?
- Wasn't sure how to handle Exceptional Entity at all, since it raises the limits for multiple attributes on top of removing the limit entirely for Depth and one of your choice, not to mention Depth isn't a stat in Chummer yet.
- Pilot Origins might need to stat in a Control Rig bonus; I checked the Control Rig entry under Cyberware.xml but didn't see any bonuses, so I left this alone.
- Redundancy adds two maximum boxes to your Core Condition Monitor, which isn't a thing in Chummer yet.

### DIFF
--- a/Chummer/data/qualities.xml
+++ b/Chummer/data/qualities.xml
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
 <chummer>
 	<version>-380</version>
 	<categories>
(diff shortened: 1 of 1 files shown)


joha4270:
I'm going to close this request for multiple reasons.

The <id> field is missing. This is a problem. The id is a [GUID](https://en.wikipedia.org/wiki/Globally_unique_identifier), and have to be present for chummer to handle stuff properly. I use [this](https://www.guidgenerator.com/online-guid-generator.aspx) tool to generate them.

Chummer don't support AI's yet. No ETA but not soon, as too multiple things are above in the priority list, so until depth is a thing, this is not going to be happening.

Additionally i can help you with some of your missing features

>No idea how to get Designer to apply its bonuses to a specific device; probably best to just add a text field for the user to enter their home device?    
Something like that, don't have a selectgear at current moment

>Wasn't sure how to handle Exceptional Entity at all, since it raises the limits for multiple attributes on top of removing the limit entirely for Depth and one of your choice, not to mention Depth isn't a stat in Chummer yet.      
You should take a look at the <selectattribute> and <specificattribute> (hint, no practical difference between no limit and 99, just keep it under 100 to not change ui layout :P)

>Pilot Origins might need to stat in a Control Rig bonus; I checked the Control Rig entry under Cyberware.xml but didn't see any bonuses, so I left this alone.     
Good :P, i don't think there is anyway for this to work without code support either

>Redundancy adds two maximum boxes to your Core Condition Monitor, which isn't a thing in Chummer yet.     
Look at what cyberlibs does, ( i don't think specifying core would break anything, but most likely we would disguise physical as core)

SydneySedai:
Ah, yeah, I meant the core monitor in specific, not the ability to add boxes in general.

PR #396, opened 2015-09-17
Fixes to Positive and Negative Quality limits

cfresquet:
Fixed the ExceedNegativeQualities option
Fixed the ExceedPositiveQualities option
Fixed the ExceedNegativeQualitesLimit option

CalculateBP() was incorrectly summing over multiple enemies - 1st enemy would be correct, 2nd enemy would add iteratively to first (1+1+2), 3rd would add in both the other two (1+1+1+2+2+3) etc.

So was ValidateCharacter()

These two functions BP cost calculation for Enemies sometimes used karma cost multiplier of qualities, sometimes using karma cost multiplier for contacts -- it is unclear which to use, but they definitely need to be consistent -- so I made a command decision use karma multiplier for contacts.

Further comment for future changes:
The entire Enemy system is a holdover from 4th ed, and doesn't actually exist in 5th.  So I didn't do a full cleanup of all the enemy code -- just the parts in the CalculateBP() and ValidateCharacter() functions.
Note: the enemy system apparently uses karma cost multiplier for qualities, so it is currently inconsistent with CalculateBP() and ValidateCharacter().  Rather than go through and change it all around, I figured I'd leave it to be done all at once when Enemies gets a cleanup pass.

### DIFF
--- a/Chummer/Character Creation/frmCreate.cs
+++ b/Chummer/Character Creation/frmCreate.cs
@@ -15505,13 +15505,12 @@ namespace Chummer
         private int CalculateBP()
         {
             int intPointsRemain =_objCharacter.BuildKarma;
-            int intPointsUsed = 0;
-            int intEnemyPoints = 0;
-            int intNegativePoints = 0;
+            int intPointsUsed = 0; // used as a running total for each section
             int intFreestyleBPMin = 0;
             int intFreestyleBP = 0;
             string strPoints = LanguageManager.Instance.GetString("String_Karma");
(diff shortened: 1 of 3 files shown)


chummer5a:
Yeah, the previous dev enabled the enemy code as a request. Thanks for cleaning it up!

cfresquet:
After thinking about it overnight, I'll add a separate Karma Multiplier option for Enemies, and finish cleaning up the Enemy system.

PR #413, opened 2015-09-23
Rewrite of skill costs to fix discounts for qualities

cfresquet:
Added correct bonuses for 
   School of Hard Knocks
   Linguist
   Tech School
   College Education

Also did some minor cleanup of frmCreate::CalculateBP() to make it easier to work on in the future.

### DIFF
--- a/Chummer/Character Creation/frmCreate.cs
+++ b/Chummer/Character Creation/frmCreate.cs
@@ -15381,8 +15381,8 @@ namespace Chummer
         /// </summary>
         private int CalculateBP()
         {
-            int intPointsRemain =_objCharacter.BuildKarma;
-            int intPointsUsed = 0; // used as a running total for each section
+            int intKarmaPointsRemain =_objCharacter.BuildKarma;
+            //int intPointsUsed = 0; // used as a running total for each section
             int intFreestyleBPMin = 0;
             int intFreestyleBP = 0;
             string strPoints = LanguageManager.Instance.GetString("String_Karma");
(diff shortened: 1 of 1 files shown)


joha4270:
Finally ;)

Too tired to get an overview of the code right now (i don't mind you renamed pointused, but it makes it harder to see bread and butter of the change (2 commits next time, pretty please with sugar on top?))

Too tired to get an idea what the code is doing right now, will take a look tomorrow. (or chummer5a might just merge while i sleep)

cfresquet:
Sorry, the renaming of intPointsUsed happened halfway through the process because I kept screwing myself up trying to track through the existing code, so I paused and did a refactor so I could more easily wrap my head around the other work.  If I'd been planning properly I'd have done a cleanup pass first, rewrite second.

joha4270:
Not blaming you, it needed to be done sometime, it just made it a little harder to grasph the changes.

Going to merge in 30 seconds, but before that i want make a few comments.
string.trim() !== string.Empty
Chummer uses null way too much for signaling for me not to get chills down my back when i see this. String.IsNullOrWhitespace is much preferred. (but in the end, personal preference)

Also, chummer uses tabs, not spaces for inclination, due to historical reasons. When you use spaces, it makes code interesting to read.

cfresquet:
I didn't write the string.trim() !== string.Empty line, so I claim ignorance. :)  There are more than a few stylistic choices in the code that I look at and just shake my head (example - I'm not a fan of starting variables with their type -- that was useful 20 years ago, but IntelliSense tooltips make it pointless), but it would be WAY more effort than it is worth to refactor.

I'll change my project preferences to tabs, I honestly didn't even notice.

PR #456, opened 2015-10-17
Added Body Sculpt power data

HaikenEdge:


joha4270:
Why is this 2 pulls?

HaikenEdge:
Did I pull it twice? I thought I pulled the powers XML once and the book XML once.

joha4270:
Exactly that.

There are no reason to have the power without the book or the book without
the power, so i don't see any reason to split it in multiple.

(but no worries, 1 or 2 buttons don't do a lot of difference, just
commenting)

HaikenEdge:
Didn't realize you could put two edits into one pull request. My bad. I just put in a request based on the structure of GitHub, where had it listed pretty separately.

joha4270:
As far as i can see, you did it in separate branches

Make all changes on one branch, then create a pull for that branch.

HaikenEdge:
I'm not even sure how to do that. I just forked the files, edited them, and they ended up in different branches.

joha4270:
You can fork files? I was under the impression you always forked an entire
repository.

Not that this is really important... so lets drop it and i'll get it merged

PR #475, opened 2015-10-27
Run Faster "Buying the Basics"

HaikenEdge:
Implemented data entry for all items except "Piece of Betel Gum" in the "Buying the Basics" section (pg 253-255) of *Run Faster* (pg 253-255). Excluded "Piece of Betel Gum" entry because the price conflicts with "Betel", which is the same item and was printed in *Chrome Flesh* with a listed price of 5¥, conflicting (and superceding, due to being a newer book) the *Run Faster* listed price of 1¥.

### DIFF
--- a/Chummer/data/armor.xml
+++ b/Chummer/data/armor.xml
@@ -1804,6 +1804,490 @@
       <source>SR5</source>
       <page>438</page>
     </armor>
+		<armor>
+			<id>76770acb-ec35-4e8b-ba8c-f757c34ca47f</id>
+			<name>One-Use Shoes</name>
+			<category>Clothing</category>
+			<armor>0</armor>
+			<armorcapacity>0</armorcapacity>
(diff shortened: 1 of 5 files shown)


cfresquet:
I thought Betel gum was 5¥ for a pack, 1¥ for a single stick?
...with no specification of how many sticks in a pack.

HaikenEdge:
Betal is listed as 5¥ per dose in *Chrome Flesh*, which I would think would be a single piece, and not an entire pack.

cfresquet:
Hmn.  You are correct, sir.  Plus it is listed as 5¥ in two different places in Chrome Flesh, while the entry in Run Faster is more of a background/flavor reference.

I'd say 5¥ should pretty clearly be the correct price.

HaikenEdge:
Hence the reasoning as to why the entry was excluded.

PR #533, opened 2015-11-24
add german pages from DataTrails Commlinks, Cyberdecks and Cyberdeck modules

masterki:
add german name of Data Trails (Datenpfade)
add german pages from DataTrails for Commlinks and Cyberdecks
correct some german page numbers
translate Cyberdeck-Modules, correct german hacking program names and
pages


Fightbackman:
I looked through everything. Seems to be correct translated.

PR #589, opened 2015-12-23
Merge remote-tracking branch 'refs/remotes/chummer5a/master'

holgerschlegel:
Some more german translations and core rulebook page informations.

To make the translation for the positive quality "Resistence for pathogens and toxins" visible, I had to update the qualities.xml file. It did not contain an id tag for that quality.

That change works for me, but I'm not 100% sure if it is correct for all situations. Please check.


holgerschlegel:
Looks like I did something wrong creating this pull request. I'm relatively new to Git and GitHub.
What I want to contribute was the second commit (Updated translations) only.

Is there a way to fix this?

PR #671, opened 2016-01-31
Changed costs of the autopicker

Keroth:
Corrected the costs of the auto picker using the german core rulebook.


chummer5a:
I appreciate  the contribution, but currently we don't take the view that the German books are errata; I would rather see this as a separate item, or an override_gear.xml file that users could enable for themselves.

PR #682, opened 2016-02-10
Initiativeboard for GM-Fight tracking

GhostWhoWalksInside:
Not sure whether this changes are needed in main branch, but as far as i understood some people might like the possibility to track Initiatives and damagevalues during a fight, so i offer my 2 cents. Most of the parts were existing before, i just did some minor changes. It´s not finished jet, as far as the Dashboard part could show a lot more information, but it´s usable and i would like some feedback at this point, which path i should take.
As far as i did not write any changelog, here a short overview:
- Disabled update, (sorry, but did not want to risk my GM would do an autoupdate and loose all changes)
- enabled GMDashboard in options menu (option was allready implemented, just deactivated)
-reworked ConditionMonitor on GMDashboard
-reduced the ammount of Skills shown for chars to just the ones with at least a value above 0
-added small spell control, showing all spells of char and allows fast spell using. 
- enhanced InitiativeControl: 
-possibility to add initiative modifiers for the whole battle, either dice or valuebased 
-option to keep initiative roll for the whole fight, to speed up fights a bit 
-damage affects initiative values for the battle (needs some tweaks maybe to get all the enhancements used for calculating the damage modifier)

so far it´s right now, so if there are any hints or wishes, just add them here.
Greets, Ghostwalker


joha4270:
Okay, not going to merge this. At least not now. 

Here is a not exhaustive list of problems i have (I only played a few minutes)
* A lot of buttons have text surrounded by '{' and '}'. Not sure it is intentional, but it mostly looks like something was not loaded probably
*  There was a file->load option, but nothing happened when i pressed it. This is going to give me about a quarter million of bug reports. 
* I crashed when i tried to roll with no skills. Might need a little more QA
* skills are being rewritten at [chummer5a:newskills](https://github.com/chummer5a/chummer5a/tree/newskills). I am slightly unhappy adding new stuff that uses the old skills.

chummer5a:
That said, we appreciate your efforts! Having the gm dashboard actually work has been in my backlog for ages now. I think for now it would be best to keep it as a separate fork

GhostWhoWalksInside:
Soka, no problem, I will keep it for myself for now.
I at least got some information what you care about.

GhostWhoWalksInside:
Maybe we can agree on an Interface for Skills, so you can tinker around with your skillrework and i can polish the dashboard ?

joha4270:
Sure. 

Well, there is a more or less finished one over at the newskills branch, it just needs buttloads of polish before i can happily call it finished.

Either you can watch for when it gets integrated into master or i might do a PR once i get bored.

Issue #717, opened 2016-02-25
Skillwires Essence not calculated properly


Blade98:
### Expected behaviour
As per SR5 456, Skillwires are Rating x 0.1 for essence therefore adding Skillwires at Rating 1 should cost 0.1 essence. Addning Skillwires at Rating 2 should cost 0.2 essence

### Actual behaviour
When adding Skillwires at Rating 1, the essence cost is reported as 0.2. When adding Skillwires at Rating 2, the essence cost is reported as 0.4

### Steps to reproduce the behaviour
Add Skillwires into the Cyberware and Bioware tab

HaikenEdge:
Corrected in https://github.com/chummer5a/chummer5a/pull/718 and awaiting merge.

tcjennings:
This bug exists again in 5.204 and current master.

chummer5a:
Not a bug now. It's in Chrome Flesh as being 0.2 per Rating.

tcjennings:
Chrome Flesh has skill wires at 0.1 per rating on page 225 and 0.2 per rating on page 78 and again on page 222. Sigh. I guess 2 votes beats 1. Thanks.

chummer5a:
[amend_skillwire_cyberware.zip](https://github.com/chummer5a/chummer5a/files/2849862/amend_skillwire_cyberware.zip)
If your GM's kosher with using the CRB version, you can use this to amend it.

PR #718, opened 2016-02-25
Skillwires Essence Correction

HaikenEdge:
Corrected from "Rating * 0.2" to "Rating * 0.1" as per the book.


joha4270:
Just for the future, you can type fix #number and github will understand it. 

"Skillwire essence correction, fix #717"

Issue #766, opened 2016-03-18
Adept powers miscalculated when improved by e.g. mentor spirit (pull request submitted)


graptis:
### Chummer Version
5.178, 5.179 tested

### Expected behaviour
Create an Adept character, add Mentor Spirit (Eagle) and an additional level of combat sense for a total of **Two**. Save the character. Load the character again. Combat sense is **Two**.

### Actual behaviour
Create an Adept character, add Mentor Spirit (Eagle) and an additional level of combat sense for a total of  **Two**. Save the character. Load the character again. Combat sense is **One**.

### Steps to reproduce the behaviour
See above.

### Helpful Fields
PowerControl.cs line 65
				if (actualRating > _objPower.FreeLevels)
replace to 
				if (actualRating > 0)
Rationale: Actual rating seems to be the number of power points spend by the user, (in the above scenario, **1**. The above conditional seems to ensure that the total does not drop below the FreeLevels, which would only happen if the actualRating dropped below 0. The bug manifests because when the user has added one level of Combat Sense, both actualRating and FreeLevels are one, causing the control to set the power level to **one** instead of **two**.

### Operating System
Windows

### Use of international language content
-

### Last known good version
unknown

### Link to affected character file
On loading the character, power level is 1. Increase to 2. Save. Load. Power level is one.
[TEMP.zip](https://github.com/chummer5a/chummer5a/files/180251/TEMP.zip)

PR #767, opened 2016-03-19
Adept Power control fix

graptis:
On the initial PowerControl loading, fixed the power level calculation to include both FreeLevels and actual PowerRating (Issue #766)

### DIFF
--- a/Chummer/Character Creation/frmCreate.cs
+++ b/Chummer/Character Creation/frmCreate.cs
@@ -13879,15 +13879,46 @@ namespace Chummer
                     {
                         if (objPower.BonusSource == objGear.InternalId)
                         {
+                            //Remove the Bonus Source since this object will not be giving a bonus
+                            objPower.BonusSource = "";
+
                             if (objPower.Free)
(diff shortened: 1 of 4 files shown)


chummer5a:
I have a vague suspicion this won't interact nicely with qi foci, I know there was a reason I didn't do it that way initially.

graptis:
Hmm... Can you suggest a few use cases that would give some coverage? I can test/tweak it to work properly with other cases, but as I am rather new to Shadowrun so I am unsure I can think of all things that might be interacting with it...

joha4270:
Add a QI focus

* For a power you don't have in any way
* For a power you do have
* For a power given to you by a mentor spirit
* For Improved Reflexes

(I haven't actually looked at this patch, but that is the possible issues i
can see)

graptis:
For your review :)
I have amended the fix - I also picked up a few different related issues (especially binding - unbinding - rebinding foci).

chummer5a:
Looks good to me, thanks for the contribution.

Issue #790, opened 2016-03-29
Cloak/Designer Cloak throwing ArgumentOutOfRangeException


skarsol:
### Chummer Version
5.179.781
### Expected behaviour
Cloak/Designer Cloak should have 1 rating for each desired armor mod. Adding mods up to that value should not have an issue.
### Actual behaviour
Cloaks with a rating higher than 1 throw errors when clicked on if an armor mod exists.
### Steps to reproduce the behaviour
Add a Cloak (Rating 2)
Add an Armor Mod to it (eg: Gel Packs)

### Error
See the end of this message for details on invoking 
just-in-time (JIT) debugging instead of this dialog box.

************** Exception Text **************
System.ArgumentOutOfRangeException: Value of '3' is not valid for 'Value'. 'Value' should be between 'Minimum' and 'Maximum'.
Parameter name: Value
   at System.Windows.Forms.NumericUpDown.set_Value(Decimal value)
   at Chummer.frmCreate.RefreshSelectedArmor() in C:\projects\chummer5a\Chummer\Character Creation\frmCreate.cs:line 17612
   at Chummer.frmCreate.treArmor_AfterSelect(Object sender, TreeViewEventArgs e) in C:\projects\chummer5a\Chummer\Character Creation\frmCreate.cs:line 12451
   at System.Windows.Forms.TreeView.OnAfterSelect(TreeViewEventArgs e)
   at System.Windows.Forms.TreeView.TvnSelected(NMTREEVIEW* nmtv)
   at System.Windows.Forms.TreeView.WmNotify(Message& m)
   at System.Windows.Forms.TreeView.WndProc(Message& m)
   at System.Windows.Forms.Control.ControlNativeWindow.OnMessage(Message& m)
   at System.Windows.Forms.Control.ControlNativeWindow.WndProc(Message& m)
   at System.Windows.Forms.NativeWindow.Callback(IntPtr hWnd, Int32 msg, IntPtr wparam, IntPtr lparam)

************** Loaded Assemblies **************
mscorlib
    Assembly Version: 4.0.0.0
    Win32 Version: 4.6.1073.0 built by: NETFXREL3STAGE
    CodeBase: file:///C:/Windows/Microsoft.NET/Framework/v4.0.30319/mscorlib.dll
----------------------------------------
Chummer5
    Assembly Version: 5.179.781.0
    Win32 Version: 5.179.781
    CodeBase: file:///C:/Users/pellis/Documents/chummer/Chummer5.exe
[...]

PR #793, opened 2016-03-30
Fix #791

### LINKED ISSUE #791: Radar Sensor not available in Sensor Array
### Chummer Version
5.179.781
### Expected behaviour
Radar Sensor (CF:81) should appear in the Sensor Functions list for Sensor Arrays.
### Actual behaviour
It does not. It appears to be only attainable as implanted Headware. This listing for this is more "legit" than the one for X-Ray (only mentioned as an option on a specific drone and has no rules that I can find) and X-Ray is included in Sensor Functions, so makes sense this should be there too.

skarsol:
Apply fix for #791. Included cost in this sensor as it's special/expensive. Most people I see include costs for sensor upgrades like Thermographic Vision on a Camera, and this is kinda/sorta similar? Especially since it has relevant Rating levels that determine performance unlike the other 0 cost ones.

(My first Pull Request, please let me know if I did anything wrong/weird.)

### DIFF
--- a/Chummer/Character Creation/frmCreate.cs
+++ b/Chummer/Character Creation/frmCreate.cs
@@ -17609,8 +17609,8 @@ namespace Chummer
 				}
 				else
 				{
-					nudArmorRating.Value = objArmor.Rating;
 					nudArmorRating.Maximum = objArmor.MaxRating;
+					nudArmorRating.Value = objArmor.Rating;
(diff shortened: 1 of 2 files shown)


skarsol:
Fixes #790. Trying to set the value in the Rating box to a number >1 after adding a Mod with no rating causes the error.

Issue #786, opened 2016-03-29
Horizon CU^3 missing professional option


skarsol:
### Chummer Version
5.179.769
### Expected behaviour
Horizon CU^3 can purchase the Professional Upgrade
### Actual behaviour
There appears to be no mention of the Professional Upgrade
Rigger 5, page 130: Professional Upgrade: Pilot increased to (4), Clearsight increased to (4), for +3,000¥

skarsol:
Fixed.

PR #794, opened 2016-03-30
786

skarsol:
Add a professional version of the CU^3. This seemed way cleaner/easier than trying to apply a purchasable mod to the base version.


skarsol:
This addresses #786

Issue #780, opened 2016-03-28
Modifying stats on drones creates an exception.


skarsol:
### Chummer Version
5.179.769 and .775

### Expected behaviour
Adding Handling (Drone) Rating 5 to a Handling 4 drone should make it Handling 5.
It shouldn't throw an error.
### Actual behaviour
It adds the Rating to the current value. (Handling becomes 9)
It throws an error.
Unhandled exception: Input string was not in a correct format.
### Steps to reproduce the behaviour
Add a drone.
Add a modification.
Select any base stat (eg: Handling (Drone))
Select any rating.
Hit OK

Blackbaud-SeanLong:
Duplicated on a DEBUG build, 5.179.781.

Same error. Exception text follows:

System.FormatException: Input string was not in a correct format.
   at System.Number.StringToNumber(String str, NumberStyles options, NumberBuffer& number, NumberFormatInfo info, Boolean parseDecimal)
   at System.Number.ParseInt32(String s, NumberStyles style, NumberFormatInfo info)
   at System.Convert.ToInt32(String value)
   at Chummer.VehicleMod.get_CalculatedSlots() in C:\projects\chummer5a\Chummer\Classes\clsEquipment.cs:line 13623
   at Chummer.Vehicle.get_SlotsUsed() in C:\projects\chummer5a\Chummer\Classes\clsEquipment.cs:line 14954
   at Chummer.frmCreate.RefreshSelectedVehicle() in C:\projects\chummer5a\Chummer\Character Creation\frmCreate.cs:line 19351
   at Chummer.frmCreate.tsVehicleAddMod_Click(Object sender, EventArgs e) in C:\projects\chummer5a\Chummer\Character Creation\frmCreate.cs:line 9108
   at System.Windows.Forms.ToolStripItem.RaiseEvent(Object key, EventArgs e)
   at System.Windows.Forms.ToolStripMenuItem.OnClick(EventArgs e)
   at System.Windows.Forms.ToolStripItem.HandleClick(EventArgs e)
   at System.Windows.Forms.ToolStripItem.HandleMouseUp(MouseEventArgs e)
   at System.Windows.Forms.ToolStripItem.FireEventInteractive(EventArgs e, ToolStripItemEventType met)
   at System.Windows.Forms.ToolStripItem.FireEvent(EventArgs e, ToolStripItemEventType met)
   at System.Windows.Forms.ToolStrip.OnMouseUp(MouseEventArgs mea)
[...]

Issue #787, opened 2016-03-29
Unable to create "Fragile" drones


skarsol:
### Chummer Version
5.179.781
### Expected behaviour
Drones should be able to add the "Fragile" upgrade
### Actual behaviour
Option doesn't exist
### Additional Info
Rigger 5, p123: Reducing the drone’s Body by 1 produces 2 Mod Points, resulting in a net gain of 1 Mod Point.
The MCT Seven show the representation of this as "Fragile (1)". (And is reflected in their 1 Body instead of the normal 2)
This brings up the side issue that Mod Points from Rigger 5 aren't addressed in Chummer, but that's something that can be worked around.

skarsol:
Fixed.

PR #797, opened 2016-03-31
Vehicle/Drone modification tracking

skarsol:
So this adds display of the available modification slots for Vehicles and Drones. Vehicles show the 6 categories of mod pools and tracks them separately. Drones just show their single pool of mod slots. The free upgrade for each stat for Drones is accounted for. Immobile adds the 2 slots correctly, but I've not coded for the 1 mod point available by reducing a Drone stat by one. I'm thinking the best way for that would be to add a Negative Vehicle Mod which reduces the stat and adds the 1 mod point. Didn't want to do that till I got some comment on this patch and that concept.

This also fixes some code that I ran into in the process which still does things by old rules (Speed/Accel having percentage mods, etc). I couldn't find any instances where those still applied, but it was breaking current functionality.

This patch also fixes issue #780 and opens the path to fixing #787 with a gear similar to the modified Immobile.


skarsol:
FYI, the drone code on this is notably lacking a switch to turn it off based on the Rigger 5.0 flag. I wasn't sure if it was best to use that flag or to create a new one (so you could use the other rules in Rigger 5, just not the drone mods rules (ie: you still want your assault rifles from SR5 :P))

chummer5a:
Given that they do present multiple options for how to handle modifying your drones in the rulebook, I'd be happiest if it were an optional rule. I've had more than a few requests that it not be something we shove down people's throats. Shouldn't be a huge deal to implement, the Options code is pretty basic and there's already the rather basic IsDrone boolean if you want to go that far. Simplest way for us to handle the mod point increases is probably going to be mod items that have a negative bonus for whatever rating they are and a negative value for their Slots equal to rating.

skarsol:
Well, this works on my test build. I'm not sure why Appveyor is failing. Looking at that commit, I don't see a merge issue. Not really sure how to investigate that.

skarsol:
Found a couple things, set up the upstream link, got all up to date, Appveyor still complaining. Advice appreciated. :)

chummer5a:
Appveyor tends to scream if it has to merge changes. Since I was doing stuff upstream with the Acceleration bug (Costs are listed as Acceleration * whatever instead of Accel), it got a bit unhappy. In theory should be fine now.

PR #823, opened 2016-04-21
Nanoware data entry.

Ternega:
Added Nanoware from CF. Duplicated some of it (drug carrier and antidote
carrier) to gear and added weaponized nanotech to gear.

### DIFF
--- a/Chummer/data/cyberware.xml
+++ b/Chummer/data/cyberware.xml
@@ -2877,7 +2877,7 @@
 		<!-- Region Hard Nanoware -->
 		<cyberware>
 			<id>9b4e4b37-8f0e-4224-82b6-158077551e1b</id>
-			<name>Anti-rad</name>
+			<name>Anti-Rad</name>
 			<category>Hard Nanoware</category>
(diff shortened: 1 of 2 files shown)


chummer5a:
Thanks for the commit! Couple of things though; 

- we prefer to have names in proper case
- the master already seems to have most of these entries; if you're replacing or moving them in the structure, could you please make sure that the IDs stay the same? It can cause some issues with item concurrency if they're not in sync. 

Looks good aside from that though.

Ternega:
There is an issue, see, only nanoware in masters cyberware.xml is the Implant Medic, for it I simply moved it keeping ID and adding forced grade to it. I didn't even realise that other ware I entered was in the files you see, because it's all listed as category Hard Nanoware in gear.xml, except gear.xml doesn't have Hard Nanoware category defined so it was inaccessible from the application itself.

I can move ID's from gear.xml to cyberware.xml and delete them from gear completely, if you prefer. But I'd like to know how you prefer it be resolved before I do so.

PR #838, opened 2016-05-11
corrected removal of lifestyle qualities

OLStefan:


joha4270:
Going to look at it when i get home in 4 hours time, then probably merging it.

If you have more discipline that i do, you can name your commits fix #number and github will close the issue once it is merged.

PR #852, opened 2016-05-25
Rewritten skills

joha4270:


joha4270:
Merged with old master, will redo

PR #873, opened 2016-05-30
some superficial translations

HaukeW:
down until gear.xml. Some strings need to be translated though.


HaukeW:
Ignore the pull request please, I am new to github and don´t know how to cancel it. The diff in de_data.xml is wrong but i realized it too late (see https://github.com/chummer5a/chummer5a/issues/874 )

PR #881, opened 2016-06-01
Corrected Ingram Smartgun X

Skarablood:
Added Gas-Vent 2 system as standard addon to Ingram Smartgun X, as per it's description in the core rulebook, p. 427.


chummer5a:
From memory this will now have an RC of 4, as integral modifications apply their bonuses now. You'll need to drop the RC back to 0 to compensate.

PR #1023, opened 2016-07-27
fixed recoil compensation for some machine guns

bodison:
Ruhrmetall SF-20 (now has Gas-Vent 3 System), Ultamax HMG-2, Shiawase Armaments Nemesis, FN MAG-5

name fixed for Ultamax MMG


chummer5a:
Ultamax looks so wrong, but whatever, it's in the book that way.

PR #1032, opened 2016-08-03
made some german translations, also added some values to de.xml that had no translation in chummer

HaukeW:
Question (I don´t know if it is important): While i was browsing en-us.xml I noticed the following pattern: 

- older entries tend to have spaces in front of the values and 
- newer entries tend to have tabs in front of the values.
- Also de_data.xml only has tabs in front of the values instead of spaces while de.xml mostly has spaces.

So: Should this in de.xml be unified to "only tabs in front of values instead of spaces"?


HaukeW:
Oh, by the way, the many commit-reverts happened because I made a typo ("stringe" instead of "strings") and thought I could just correct it (in the sense of: make it disappear) by reverting it.

Didn´t work. Obviously.

chummer5a:
For the most part space/tabs doesn't matter, Git normalises everything to spaces anyway. I have a feeling that the translator application uses spaces, but it's been a while since I cracked that code open. My personal preference is tabs, but I've never bothered to codify a style guide. Use whatever you want, as long as you're not changing a whole file for the sake of normalisation.

HaukeW:
Ok, thanks!

PR #1178, opened 2016-10-26
Updated Thrifty to match book description.

HaikenEdge:
Rules text: "Reduce the points available by 2 and the monthly cost by 1,000 nuyen. Only available for the traveler lifestyle."
Not sure how to set requirement for Traveler lifestyle, but at least now the NQ will function.


chummer5a:
Implemented separately in order to manage the Traveler requirement.

PR #1268, opened 2016-12-03
Drone downgrades

OLStefan:


Chummer/Backend/Equipment/Vehicle.cs
```diff
@@ -1383,14 +1382,27 @@ public int DroneModSlots
 		{
 			get
 			{
-				int intDroneModSlots = _intDroneModSlots;
+				int intDroneModSlots = _intDroneModSlots;
+			    bool blnDowngraded = false;
```
crazymykl: Your indentation seems different than the rest of the file.

PR #1523, opened 2017-03-10
updated German translation

Noraser:
Included:
- corrected German Books naming (see Issue threat)
- all translated Qualitys from Run Faster (Schattenläufer) to German including the page transformation to the german books.
Base is the de_data from 191.27

[de_data 191.27.zip](https://github.com/chummer5a/chummer5a/files/834327/de_data.191.27.zip)


Noraser:
obviously I got it wrong, sorry...

PR #1613, opened 2017-05-04
update spells.xml

dashifen:
added ids, fixes to range (e.g. LOS(A) becomes LOS (A)) and duration (1 becomes I).  note: I'm reasonably sure that the GUIDs I added are unique, but I don't have a good way to double check.  If you've a good tool for that, it's probably worth checking.


chummer5a:
Chummer will automagically check for non-unique GUIDs in a given category on load.

PR #1614, opened 2017-05-04
Resist Pain DV Fix

dashifen:
Resist Pain's DV was listed as simply -4 and not F-4.  I've fixed in this one.


dashifen:
Crud - forgot to make a new branch with its own commit.  Do I need to re-do or can you manage?

PR #1758, opened 2017-06-28
[WIP] Added Social Apperance Anxiety ratings 2 and 3

mudge6:
- [x] Added ranks
- [x] Added GUIDs

Im not sure how to add GUIDs, if someone more experienced could help, that would be nice.

### DIFF
--- a/Chummer/data/qualities.xml
+++ b/Chummer/data/qualities.xml
@@ -12067,7 +12067,7 @@
 			<page>152</page>
 		</quality>
     <quality>
-      <id></id>
+      <id>75888c21-ab20-4133-8d1e-c47809b9dc99</id>
       <name>Social Appearance Anxiety (Rating 2)</name>
       <karma>-6</karma>
       <category>Negative</category>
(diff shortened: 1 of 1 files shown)


chummer5a:
Use something like https://www.guidgenerator.com. Most IDEs have something to generate them as well. I generally use the guid generator in http://phdesign.com.au/npptoolbucket with my notepad++ installation.

mudge6:
@chummer5a alright just added GUIDs, anything need to be changed or added?

chummer5a:
Hrm. You can probably implement the penalty with the skillcategory improvement type, see if this works
`        <skillcategory>
          <name>Social Active</name>
          <bonus>-1</bonus>
        </skillcategory>`

mudge6:
@chummer5a personally i think that should be left out since its not active all of the time, unless you know of a way to have a modified dice pool added to the number

chummer5a:
Oh right, it's conditional on not being pretty. Yeah, that's fine as is then.

PR #1794, opened 2017-07-11
[WIP] Fixes #1792: Spending SP on Knowledge Skills causes false negative validation

### LINKED ISSUE #1792: Spending SP on Knowledge Skills causes false negative validation.
### Summary

Issue is one of design. Assumptions made in $(git rev-parse --show-toplevel)/Backend/Skills/SkillsSection.cs - public int KnowledgeSkillPointsUsed (Line 428)

`get { return KnowledgeSkills.Sum(x => x.CurrentSpCost()); }`

States that the number of Knowledge Skill Points that have been used are equivalent to the number of Knowledge Skill Ranks total, which is untrue. This problem is further exacerbated by the fact that public int SkillPoints (line 436) is doing too much work with:

`if (KnowledgeSkillPointsUsed > KnowledgeSkillPoints)
			work -= KnowledgeSkillPoints - KnowledgeSkillPointsUsed;`

Perhaps calls should be made to in place of this calculation as well as 

`KnowledgeSkills.Sum(x => x.CurrentSpCost());` 

in order to make their results accessible separately such that KnowledgeSkillPointsUsed could accurately reflect that the number of knowledge skillpoints used as being the sum of knowledge ranks sans the number of skill points used to buy knowledge skillpoints. 

### Environment
Chummer Version: 5.193.0 (If that's incorrect it's whatever the latest is, I pulled and built from source).
Environment: Windows 10, 64 bit.
Runtime: .NET 4.5.2 / Also tested with 4.5.1

### Expected behavior
Upon spending SP on Knowledge Skills it decrements SP but not Knowledge Skills (Ex: I have 0 knowledge skill points and 1 skill point, I rank up a knowledge skill. It removes one SP and I'm left with 0, knowledge skill points remains at zero).

### Actual behavior
Spending SP on Knowledge Skills decrements both SP and Knowledge Skill Points (Ex: I have 0 knowledge skill points and 1 skill point, I rank up a knowledge skill. It removes one SP and I'm left with 0, knowledge skill points go to -1 which causes a fail in validation).

### Steps to reproduce the behavior
1. Create a new character.
2. Spend all Knowledge Skill Points.
3. Attempt to buy more Knowledge Skills. 
4. Click Special -> Validate character.

### Link to affected character file
Test Character File provided in attachments

[Knowledge Skill Bug Sheet.zip](https://github.com/chummer5a/chummer5a/files/1137730/Knowledge.Skill.Bug.Sheet.zip)

Xenryusho:
## Summary of changes

I extracted logic such that the the sum of skill ranks, and the number of skill points spent on knowledge skills could be used individually. My functional change is on line 430:
`get { return KnowledgeSkillRanksSum - SkillPointsSpentOnKnoskills; }`
where previously it read
`get { return KnowledgeSkills.Sum(x => x.CurrentSpCost()); }`
which is functionally/logically equivalent to the new KnowledgeSkillRanksSum accessor.

The code in SkillPointsSpentOnKnoskills is directly taken from SkillPoints as it was written.

## Concerns

- Am I missing any documentation that should be updated as well?
- Any specific testing that might be prudent?
- Naming for accessors KnowledgeSkillRanksSum and SkillPointsSpentOnKnoskills
- I'm not sure how to check the compliance of my line endings. I noticed a commit was recently made to fix endings and I don't want to add nonconforming code.
- It seems that the rest of the file is using tabs, my editor is set to emits four whitespaces which fits with the recommended style; is this difference a concern?
- Should accessors added be public? I see no reason this information shouldn't be exposed to other systems, and I feel like there might be a legitimate use for them to have this information available but unsure. 

## TODO:
- [x] Pull out logic for calculating Knowledge Skill Rank Sum.
- [x] Pull out logic for calculating number of Skill Points spent on Knowledge Skills.
- [x] Minor refactoring to Skill Points to use extracted logic.
- [x] Functional changes to KnowledgeSkillPointsUsed.
- [x] Testing that issue worked against has been resolved.
- [x] Testing compatibility with old characters / builds.
- - [x] Old characters still in creation mode load correctly.
- - [x] Old characters in career mode load correctly.
- - [x] Able to save new characters in creation mode and subsequently reload correctly.
- - [x] Able to save new characters in career mode and subsequently reload correctly.
- - [x] Able to load characters made with changed version in current chummer version.
- - [x] Regression testing? *
- [x] Update documentation **

\* I grep'ed through the solution for uses of the affected call(s) to attempt to ensure I wasn't breaking something somewhere else, didn't find anything relevant and wasn't able to see any problems by playing around with the affected systems/interfaces. 

\*\* Don't think there's any documentation that needs to be changed with what I did, changelog looks like a vdd but I only see it updated when the data files are changed.


Xenryusho:
Played around with it some more, unless there's specific regression tests that are typically done I don't see any problems so far as breaking something.

PR #1910, opened 2017-09-23
Mentor Spirit Fixes, Misc. Minor Fixes

DelnarErsike:
Application Changes:

- Fixed archaic handling of how restrictions are applied to bonus Adept powers by adding a "bonus override" portion to bonuses that add specific Adept powers with certain restrictions (e.g. specifically free levels in Attribute Boost AGI instead of in any Attribute Boost).
- Fixed a bug where bonuses granting free levels of certain adept powers, e.g. Improved Ability, would not actually have their effect applied, and would also not have their effect properly removed if the source of the bonus was removed.
- Made minor performance optimization to code that constructs the dicepool tooltips for skills.
- The "Select Skill" form can now process multiple skill group inclusions, skill group exclusions, and skill category inclusions the same way it does for multiple skill inclusions (the old way of handling multiple skill category inclusions still works). It can also handle multiple filters at once, e.g. list only Physical Skills that are linked to STR, or list all Social Skills excluding the ones in the Acting skill group.

Data Changes:

- Updated all mentor spirits whose free adept powers have a restricted selection (e.g. Firebringer, Wise Warrior, Seducer) so that those restrictions properly work.
- Corrected the text of Seducer mentor spirit's adept bonus to match the one found in the core rulebook.
- Corrected the page number references of Forbidden Arcana mentor spirits.
- Spider (Alt) now properly gives 2 levels of Spirit Claw instead of 1.
- Changed the name of "Great Mother" mentor spirit from SASS to "Great Mother (SASS)" to differentiate it from an identically named mentor spirit from Forbidden Arcana.

### DIFF
--- a/Chummer/Backend/Powers/Power.cs
+++ b/Chummer/Backend/Powers/Power.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.ComponentModel;
 using System.Globalization;
@@ -90,7 +90,7 @@ namespace Chummer
             CharacterObject.SourceProcess(_strSource);
         }
 
-        public void Create(XmlNode objNode, ImprovementManager objImprovementManager, int intRating = 1)
+        public void Create(XmlNode objNode, ImprovementManager objImprovementManager, int intRating = 1, XmlNode objBonusNodeOverride = null)
         {
             Name = objNode["name"].InnerText;
             _sourceID = Guid.Parse(objNode["id"].InnerText);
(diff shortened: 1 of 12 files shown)


cfresquet:
How can Spider (Alt) give 2 levels of Spirit Claw?   Spirit Claw doesn't have levels.

DelnarErsike:
*Sigh* Catalyst is at it again. You're right, Spirit Claw doesn't have levels, but the RAW description for Spider Alt in Forbidden Arcana says two levels of Spirit Claw. If/When they errata it, we can change it back, Chummer5 should be able to handle it regardless.

PR #1924, opened 2017-09-29
Update Chinese Language

Youneko17:
Well, here' s a three-member group working on translating Chummer5 into Chinese Simplified. We have finished the translation of the SR5 core book, and some parts of RF, R&G, DT, SG and CA.
To be honest, I have totally no clue about how GitHub works and I have already made a mistake. So I hope this time it could work.
Besides, we found some text would not be translated, though we found the key in en-us.xml and added translation. There's also some texts with no key or id, though they definitely should be translated, i.e. the talent in Priority mode. We'll make a list for them.


DelnarErsike:
The structure looks fine to me. You'll get the hang of git and GitHub eventually. Note that some of the latest Nightly updates added quite a few new strings, and it's plausible that more will be added before it gets packaged into a Release version.

Re: unlocalizable strings, if your group could assemble a list and submit it as an Issue, that would make it easier for us to find and fix them. Strings that cannot be properly localized affect every localization, so we would definitely want to rectify any of those.

If you encounter any problems or incompatibilities with East Asian script support in the UI, it would be great if you could let us know about those as well.

PR #2056, opened 2017-11-15
Fixing actions.xml mistakes

apathy-dude:
Adding closing tags for boost and fixing typo for limit


DelnarErsike:
Note that actions.xml isn't currently used in Chummer5 at all, it's just a WIP setup for a possible system in the future. The corrections are welcome, but they're not going to change anything in available through the main program.

PR #2088, opened 2017-11-22
Missing degrees of Illness Quality

Lochabar:
The Illness Negative Quality (Bullets and Bandages p12) shows three levels  for the Quality, unless I've missed an errata somewhere. Currently the Qualities data file list only one level. Here are the three levels.


DelnarErsike:
Illness is a leveled quality, so it is treated like all other leveled qualities. See the node in the original that says `<limit>3</limit>'? That indicates the number of times the quality can be taken, which is 3 for the 3 levels.

Issue #2106, opened 2017-11-27
Dark Terrors Sourcebook


chummer5a:
Content to implement:
- [x] Insect 'metavariant' critters as subtypes of existing insect spirits. Pages 22-32.
- [x] Free Insect Spirit quality. Requires point buy only, broadly uninteresting from a code perspective beyond that. Add as a quality in the same manner as Infected. Pages 32-33.
- [x] New critter powers for shedim and types. Pages 49-51
- [x] CFD Treament gear, page 83.
- [x] New technomancer Echo, page 85.
- [x] New critter powers for Infected. Requires limitation based on infected quality taken UNLESS certain other qualities are present. Pages 155-163.
- [x] New qualities for Infected. Pages 163-164.
- [x] New tradition and metamagics. Page 165.

PR #2167, opened 2017-12-12
Added dark terror infected qualities

mudge6:
Not sure if I did this right, if nothing is wrong, it should check off one of the boxes in #2106


chummer5a:
Doesn't wildcard chimera it need to add a negative quality as well?

PR #2291, opened 2018-01-12
Fixed error in string formating in priority window

luizbgomide:


luizbgomide:
I'm trying to clear my previous commits before the pull request without sucess, Only the last 2 should have been listed on the pull request.

Issue #2317, opened 2018-01-16
Very slow when creating a new character using the priority selection window


luizbgomide:
### Environment
Chummer Version: 1.196.46

### Actual behaviour
Chummer is painfully slow when creating a new character using the priority system or sum to ten. It uses a lot of CPU and memory when doing so.

luizbgomide:
They are slow due to this call `CommonFunctions.GetTextFromPDF`

PR #2318, opened 2018-01-16
Commented problematic code regarding issue #2317. 

luizbgomide:
Code doesn't work correctly and is making character creation painfully slow.
Issue #2317

### DIFF
--- a/Chummer/Classes/clsUnique.cs
+++ b/Chummer/Classes/clsUnique.cs
@@ -341,14 +341,15 @@ namespace Chummer
                 }
             }
 
-            if (string.IsNullOrEmpty(_strNotes))
-            {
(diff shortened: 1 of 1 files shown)


DelnarErsike:
This pull request has been made irrelevant with commit 9b06ea6cd761a51957b194e4533e9941a98e355b.

A deeper dive into the profiling data of the affected method showed that the needle-in-haystack search actually has an extremely minimal impact on performance. The main bottleneck was the PdfReader object's constructor, with a secondary, minor bottleneck caused by the page-to-text parser. Changing the code to cache the PdfReader object of a sourcebook once it has been constructed instead of constantly disposing and reconstructing it dramatically cuts down the overall performance cost to the point where I think it's no longer necessary to comment it out.

PR #2804, opened 2018-06-16
Spelling Change:  Cozenage should be Cozenge

dashifen:
Cozenage on line 1037 should be Cozenge.


dashifen:
Wait -- maybe errata?   Cozenage appears to be a real word....  Investigating further.

dashifen:
Yup, looks like the misspelling is actually in critters.xml

PR #2805, opened 2018-06-16
Fixes to critters.xml

dashifen:
- The technocritter with a power originally listed as Cozenge should be Cozenage.
- There was one SR5 critter with the Shadowing skill I've corrected to Sneaking
- There was one critter with "Exotic Range Weapon" instead of "Exotic Ranged Weapon."


dashifen:
Found one more spelling fix:  Natural Weapons should be just Natural Weapon.

Issue #1350, opened 2017-01-19
Adept power: Commanding Voice


Cthullik:
(Please fill out the below as accurately as possible. More detail makes it easier for us to find the problems and fix them.)

### Environment
Chummer Version:5.189.0
Environment: Win10
Runtime: (If you don't know this, it is probably .NET 4.5.1)

### Expected behaviour
There are 2 different abilities named "Commanding Voice", the ability from Street Grimoire is not given.

### Actual behaviour
Only the ability from Stolen Souls is in list.

### Steps to reproduce the behaviour
(Enter text here)

### Link to affected character file
(Enter text here)

bodison:
There are errata for street grimoire. "Commanding Voice" from Stolen Souls stands, "Commanding Voice from SG is renamed "Authoritative Tone".

chummer5a:
Yup. Not a chummer issue.

Issue #112, opened 2015-03-17
Adept Power not using Errata for Street Grimoire


chummer5a:
<a href="https://github.com/GoogleCodeExporter"><img src="https://avatars.githubusercontent.com/u/9614759?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [GoogleCodeExporter](https://github.com/GoogleCodeExporter)**
_Tuesday Mar 17, 2015 at 08:56 GMT_
_Originally opened as https://github.com/chummer5a/chummer5/issues/38_

----

```
What steps will reproduce the problem?

1.Try to select the adept power Authoritative Tone or Commanding Voice.

What is the expected output? What do you see instead?

Commanding Voice Adept power has wrong cost. The stolen souls version for 1 
power point is supposed to be used according to Street Grimoire errata not the 
0.25 with 3 levels max version. Also errata introduced new power called 
Authoritative Tone that needs input into Chummer5 as it is currently 
unavailable.

What version of the product are you using? On what operating system?

Chummer 5 most recent version as of 1/17/2015

Please provide any additional information below.

Street Grimoire errata introduced changes to the adept power that seem to have 
been overlooked. On page 4 of the Errata it will explain what I am trying to 
point out. 

To quote it. "P. 170, COMMANDING VOICE CHANGE
The version of Commanding Voice on p. 185, Stolen Souls,
stands. This power is changed to Authoritative Tone, costs
0.50 PP per level (max 3), and is described as follows:
Adepts with this power can speak in a way that immediately
makes people want to believe them and follow their orders. For
every level of this power, an adept gains a +1 die to any Opposed
Social Skill Test that they initiate (normal limits apply)."

Stolen Souls version from page 185 and 191 is right here. 
"COMMANDING VOICE
Cost: 1 power point
This power channels the adept’s magic into his voice
to enhance the modulation and pitch, subliminally influencing
the actions of any listeners. The adept takes a
Complex Action to give a simple but forceful command
(five words or less) to the target, making an Opposed
Test with Leadership + Charisma against the target(s)
Willpower + Intuition. If the adept succeeds in the test,
the target uses his next action to either carry out the
command or stands confused (gamemaster’s choice,
but the more net hits achieved the more likely it is for the
target to obey the adept’s command). Such commands
carry no weight beyond the immediate impetus, and
the affected characters will quickly reassert their wits,
returning to their original course of action. If multiple individuals
are targeted, use the largest dice pool among
[...]

PR #2810, opened 2018-06-17
Include pre-errata Commanding Voice power

neck-beard:
This is related to issue #1350 and also #112. 

I have a player using the original Commanding Voice adept power from Shadow Grimoire who can't use it in Chummer because it was later changed/removed in some errata. 

So I'm proposing with this PR to re-add in the original power printed in SG, with the title: "Commanding Voice (pre-errata)", including the original bonuses and page reference.


chummer5a:
Use an amend entry to override it for your specific campaign.

PR #3076, opened 2018-10-02
Bugfix/3067 knowledge skill cursor

### LINKED ISSUE #3067: Knowledge Skill: Autocomplete changes cursor
**Describe the bug**
The Autocomplete function for knowledge skills changes the cursor position. This will cause the input to be messed up. This seems to be a specific problem with the German version (could not reproduce it with the English version).

**To Reproduce**
1. Open Chummer5 with the German localization.
2. Switch to Skills
3. Add a new Knowledge skill
4. Type in "Militär"

**Expected behavior**
The text should be shown as typed in. The autocomplete should not replace entries automatically (like changing "Schieber" to "Fixer" without the posibillity to reverse that).

**Screenshots**
knowledgeskillbug

**Character File**
-

 - OS: Windows 10 Pro 64bit (German)
 - Version: 5.202.0

**Additional context**
-

handcraftedsource:
Fixes #3067 

(This is my second contribution to Github. So please be patient.)

The Cursor switch seems to be a general issue with Winforms ComboBoxes + DataBinding. The Knowledgeskill Combo in particular had the Problem that the user input was bound to the WriteableName property. There the new value was translated back to english (so the problem occoured when a skill matched) and written into the Name property. The getter of WriteableName returns the DisplayName which translates the Name. As far as I understand the issue the propagation of multiple PropertyChanged events during this process changed the Text property of the ComboBox multiple times.
So this PR is a only partial fix because the DataBinding sometimes still changes the cursor position. But I think its an major Improvement.

Hope the changes are still useful.


DelnarErsike:
Have you tested your code to make sure it works? Looking at the source, there are a few lines that look problematic to me (e.g. in the setter for `WriteableName`, you get the English name of the new value and then do nothing with it, and in `GetDefaultLanguageSkillName`, you perform an expensive `GetSkillTranslationNode` call and then just ignore its result if the the current UI language is English).

handcraftedsource:
Of course I tested the code. But you're right. In the case that the default language is executed, reading the translation was of course unnecessary.

PR #3129, opened 2018-10-19
This Commit adds a Unit-Test Project to Chummer, changes most of the …

ArchonMegalon:
…Application.StartupPath calls (because the path differs, the executing Assembly is actually the MSTestPlatform, not the Chummer-Assembly) and adds on single testfunction, that loads a bunch of characters. Just to show that it works.

It say's "can't automatically merge", but i really thought I worked from the most recent master-branch. Why is that?


ArchonMegalon:
Ok, I found out how to resolve this... :)

ArchonMegalon:
Also, couple of odd errors. It's prompting for missing language file strings and then throws the below exception:

That probably happens, because the Utils.IsInUnitTest does not return true for you. Maybe you are using a different runtime/Visual Studio version than I and so the AssemblyName is a different one?

Could you add a Debugger break there and check, which Assemblies are loaded for your installation there and choose another (or multiple) name(s) than I?

Of course we can change the test-data folder and we should remove any 0-byte files from there as well.
Will check these things on Monday.

chummer5a:
Updated PR works, but it won't let me merge it. Not sure of root cause.

PR #3165, opened 2018-10-30
Hamburg-gears, please read the comments before merging

Ramesch:
I tried to create the basis for the Hamburg-source book (German-exclusive?).

Since this is my first attempt to add new gear, I don't know if I did it right. So I would like to ask some questions, which should be checked before merging this request:
1. Can I simply generate and use new IDs as long as I don't find any duplicates?
2. Since I don't have any code for the Hamburg rules (like SR5 for the core rule book etc.), I chose 'Hamburg'. If someone knows the right code, please change it or let me know.
3. Due to the lack of English names for the gear, I just put the German names into armor.xml, vehicles.xml and weapons.xml. Is that okay?

If my attempts fit the rules about adding new equipment, I would add the rest from the Hamburg source book.

After the new equipment has been chosen, it should look like this:

armor

vehicle

weapon

### DIFF
--- a/Chummer/data/armor.xml
+++ b/Chummer/data/armor.xml
@@ -3069,6 +3069,52 @@
       <page>70</page>
     </armor>
     <!-- End Region -->
+    <!-- Region Hamburg -->
+    <armor>
+      <id>4b16a172-b328-4717-b632-39a3d07eeb41</id>
+      <name>FleetHead Kappe</name>
+      <category>Armor</category>
+      <armor>0</armor>
+      <armorcapacity>3</armorcapacity>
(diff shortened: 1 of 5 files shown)


chummer5a:
1. Yep, just generate a new UUID for each object. When you launch, Chummer runs all the IDs through a check for duplicates and sulks at you if it finds any.
2. Generally we go with 2- or 3- character codes for books instead of full words, German books generally get a G suffix. No real reason for it, just quicker than typing it out every time.
3. Lack of English-translated names is fine. In the unlikely event that the book gets translated to English it'll have to get ported into de_data, but it's fine as is.

PR #3209, opened 2018-11-14
Paragons from Kill Code

Ramesch:
Missing Paragons added to German translation as listed when chose 'verify data file'.

As I don't know what paragons 'Alias', 'Abort', 'Cryptome', 'Disinformation', 'Echelon' and 'Flow' are, I left them, but they are shown as mentors with 'noentry'. Maybe someone can confirm they can be deleted or should be left?


Ramesch:
'Alias', 'Abort', 'Cryptome', 'Disinformation', 'Echelon' and 'Flow' originate from the SR4 rulebook 'Unwired' and therefore do not belong to the SR5 system. They are also not part of 'Paragons.xml' (anymore), so a translation is no longer possible anyway. They have therefore been removed.

The adjustments can therefore be adopted.

Issue #3254, opened 2018-12-03
Missing Street Lethal "CorpSec Arsenal" Content


guildsbounty:
A few items from the Street Lethal book are missing...

Boats: 
Maersk Shipyards Wavecutter MPAC

Weapons
Red Samurai Katana
Saeder-Krupp/Onotari Arms Silent Claw Stiletto
Wuxing-Armtech PTL-02

Ammo
MCT/Winchester-Howe Hornet Direct-Fire Mini-grenade

Ramesch:
If everything works fine, boat and weapons should be included in the next Chummer5a-nightly. I haven't figured out how to add the ammo yet, but I'll keep trying.

PR #3277, opened 2018-12-11
Added MCT/Winchester-Grenade from SL

Ramesch:
Added missing Minigrenade from SL (should finally fix https://github.com/chummer5a/chummer5a/issues/3254 since the three weapons have already been added with https://github.com/chummer5a/chummer5a/pull/3248/commits/9e2623eab9670f1264ce6ebc91572d2792d72b86 )

### DIFF
--- a/Chummer/data/gear.xml
+++ b/Chummer/data/gear.xml
@@ -17908,7 +17908,6 @@
       <cost>5000</cost>
     </gear>
     <!-- End Region -->
-    <!-- End Region -->
     <!-- Region Street Lethal -->
     <gear>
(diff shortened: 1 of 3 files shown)


chummer5a:
Why no DV or AP?

Ramesch:
Sorry, I put the text to the commit. So again:

Since the grenade was already partially added, I had problems understanding how to add a grenade (this was my first ammo I added) and inserting it correctly (grenades have to be created twice, which is why I did not understand at first, it was not visible). When I finally had the expected result in the folder of the last Chummer-nightly, I simply forgot to move it to the Github folder to upload.

Sorry about that. Now it should hopefully work as expected.

PR #3278, opened 2018-12-14
Amend quality costs of RF (please read before merging)

Ramesch:
The karma costs for the qualities 'Dermal Alteration (Rhino Hide)' and 'Functional Tail (Prehensil)' were adjusted to the changed values according to the English rule book RF 2nd edition / Errata (see https://github.com/chummer5a/chummer5a/commit/13daa759b41f8561923cbf0f9a5d26e4da5261be).

According to the German rule book RF 2nd edition the karma costs still remain at the earlier values. So I tried to replace the values with a customrule.

rules

Unfortunately the customrule does not work as expected. Can someone have a look at this before adopting it and maybe give me a short note about what I did wrong?


chummer5a:
Naming convention's wrong on the XML file. Amends need to begin with 'amend_' and end (for qualities) with '_qualities'. In the middle you can have whatever, so the proper name would be amend_cost_qualities.xml, or just amend_qualities.xml since the folder name is descriptive enough for our purposes.

Ramesch:
Thank you!! Now it works as expected.

PR #3291, opened 2018-12-19
Tree sorting

### LINKED ISSUE #2773: Allow/automatically sort child items
### Environment
Chummer Version: 5.200.24

### Expected behaviour
Users should either be able to drag treenode items (Gear, Armour, Weapons, etc.) into a custom sort order, or automatically sort alphabetically when added. For the latter, can just use SortCustom method. Former would require some kind of ordering token that would persist through saves and loads.

Gimbalhawk:
Fixes #2773. Sort order of trees is preserved when loading a character. Also added a new hotkey, Alt+Up/Down that moves a selected tree node up or down among it's sibling tree nodes.

Is there any documentation of hotkeys I should add the new command to?


Gimbalhawk:
Whoops, it pulled in some commits I had previously committed. You can tell that git isn't my preferred source control

PR #3308, opened 2018-12-26
Adjustment of data values from RG

Ramesch:
Corrected data from 'Run and Gun' to the data given in the rule book (2nd edition).

### DIFF
--- a/Chummer/data/armor.xml
+++ b/Chummer/data/armor.xml
@@ -220,7 +220,7 @@
       <category>High-Fashion Armor Clothing</category>
       <armor>8</armor>
       <armorcapacity>4</armorcapacity>
-      <avail>8</avail>
+      <avail>10</avail>
       <cost>2500</cost>
       <bonus>
         <limitmodifier>
@@ -238,7 +238,7 @@
       <category>High-Fashion Armor Clothing</category>
       <armor>8</armor>
       <armorcapacity>4</armorcapacity>
-      <avail>8</avail>
+      <avail>10</avail>
       <cost>2500</cost>
       <bonus>
         <limitmodifier>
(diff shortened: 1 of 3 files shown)


chummer5a:
Changing the Thunderstruck seems to be an reversal of the errata; I don't see any word of an updated PDF anywhere.

Ramesch:
The weapon 'Thunderstruck' actually has 24F (instead of ‘corrected‘ 12F).

I prepared the XML files in a tabular form in Excel and then sorted them by source book and page number. After that I linked the data with the German language file to see the names and page numbers with the (German) data. But I already noticed that I made mistakes when preparing the data sources. I have therefore added 'in progress' so that the information is not yet taken over.

However, some information in the current XML files may be incorrect. But I will rebuild the linked data and then check the data in Chummer5a again.

Since I have almost all rule books only in German, the link to the German translation helps me to find the values more quickly.

But is it generally okay that I verify the data? Or do people who are closer to the rule data take care of it?

Ramesch:
Ok, I have fixed my previous mistakes and I am through with the weapons from 'Run and Gun'.

However, I have found the following differences, for which I am not sure whether they are differences between German and English rules or errors in the XML data:
- Nitama Sporter (RG 31): ammo 18(c) - in German rule book it is ‘5(m)‘
- Terracotta Arms AM-47 (RG 38): rc 1 - in German rule book it is ‘(2)‘, so it should be ‘0‘

So either one of it is wrong or I would set a German custom rule. Can someone check / confirm this?

I would like to avoid the additional expenses for English books if the differences can be clarified here.

Ramesch:
After the table links worked, I first checked the weapon data. I will submit the adjustments in the XML file in the next few hours.

chummer5a:
Ares Giantslayer Slingshot (RG 23): Ammo is '-' in the English as well. Probably needs to be treated as 1(ml) for it to work properly in Chummer. 
Nitama Sporter (RG 31): ammo 18(c) - says 18(c) in the English version. 
Terracotta Arms AM-47 (RG 38): rc 1 - Shows as 1(3) in the English version.

PR #3311, opened 2018-12-27
Sinners WIP update

ArchonMegalon:
This is a first merge for a Plugin-Framework. It can be enabled in the Options-Dialog. Currently there is only one PlugIn in Development, which can already be loaded, although it is not very usable at the moment.

It basically only allows Chummerfiles to be stored and shared online.


ArchonMegalon:
There are some big changes in there. Most notably I moved some "intelligence" from the UI, which previously handled basically all the objects, back to the objects themselves via EventHandlers. That way a PlugIn could register functions to be executed at crucial moments like loading, saving or User-Interactions like mouseclicks.

ArchonMegalon:
Wow, that code-review from GitHub is a nice feature. But actually sadly I can't edit the merge-request now.
The only thing I actually WOULD have edited is the ChummerHub/appsettings.json (remove the ",").

The Workdir.... well... I think it should stay relative. But that's up to everyone's opinion.

I blatantly refuse to edit the ChummerHub/wwwroot/lib/jquery-validation-unobtrusive/jquery.validate.unobtrusive.js That's directly from the template from Microsoft.

chummer5a:
Not sure if this is just an issue on my end, but the options menu isn't showing any controls.

PR #3327, opened 2019-01-02
Sinners merge

ArchonMegalon:


ArchonMegalon:
I really don't know what's the conflict in this file. I guess that's due to my limited understanding of git. Please just USE THE ORIGINAL ONE. I (think I) have not changed the frmCareer.Designer.cs at all (I did change the frmCareer.cs, but that does not show up as conflict here).

Thanks and sorry for the inconvenience.

ArchonMegalon:
That's a good hint with the editorconfig. Actually my IDE should support it (https://docs.microsoft.com/en-us/visualstudio/ide/create-portable-custom-editor-options?view=vs-2017) but I couldn't find any solution-files or and file called editorconfig in the solution. It DID exists in the Directory dough. 

I tried to add the config to the solution and push again.

PR #3332, opened 2019-01-03
Sinners

ArchonMegalon:
I had to delete my fork and recreate it, but finally I think I got it. At least it SHOULD be able to merge now.
Keep in mind that this is very much a work in progress. For example, you can't log out after you successfully logged in in the client (which is - well - not really a problem to test this whole thing).

To enable the plug in you have to enable the Plugin-Framework in the Options and enable the specific Sinner-Plugin in the than showing Listbox.

Also make sure, that the ChummerHub.Client.dll is in the Plugins-Directory (should be working if you run this in Visual Studio and set the active project to ChummerHub.Client).


ArchonMegalon:
Anything I can do to help here? Will be back at my Dev-Workplace at Monday and could pick up some things.

chummer5a:
Just testing out the functionality before I pull it. 

Upload doesn't seem to be working, though it could be due to my internet being garbage. 

Is there a reason we need to load and render the create/career forms to do the upload? I'd prefer to be able to keep that from showing up if possible, if there's functionality buried in the create and career forms that makes it necessary let me know and I'll see about pulling it out into the character base class.

ArchonMegalon:
> Is there a reason we need to load and render the create/career forms to do the upload?

Actually I think it's not necessary to do that right now. If you look at the frmCharacterCache Class (or better, the TreeNodes generated by the Client if some files are already uploaded), Saving and Downloading already workes (or maybe worked on my Dev-PC). Of course the Char is already open when saving him that way...

Can't get to the source right now, but if there is any dependency I will remove it.

ArchonMegalon:
I also added a changelog: 

Future Build 1.0.0.1

WebService Changes:

- wiped the Database clean (because it WAS really messed up)
- added function GetDeleteAndRecreateDb() for cases where it's really messed up
- added basic validation in the Post WebService function in SinnersController
- changed LastChange (Sinner-Property) from DateTime? to DateTime

Application Changes:

- SinnerId is taken into account after the first POST to the WebService
- moved Uploadfunction into CharacterExtended
- added a ToolTip to the Online-Treenode in case something goes wrong with the ChummerHub-Query
- reworked Multi-Upload Button in Options-Dialog
- Fixed NullReferenceException when no spell was selected in the Sinner-Search dialog.
- added Changelog ;)

PR #3460, opened 2019-02-28
Upload POST should work and first draft for groups

ArchonMegalon:
See the changelog.txt
It's more about getting some users to test it, so I can stabilize the whole upload/authorization part, while parallel working on the groups.


ArchonMegalon:
I really like the Plugin System. As you can see i only made a very minor change to 1(!) file in the main branch - the rest is all in the plugin. 

Should make it easier for you to decide, if you want to merge this... xD

PR #3465, opened 2019-03-04
Groups are implemented and I found the settings to load dlls despite of UAC

ArchonMegalon:


ArchonMegalon:
Thank you for waiting with the new milestone. I think this version should be stable! :)

PR #3468, opened 2019-03-05
Now the actual bughunting begins: - after the Post of the metadata, now the actual file is uploaded

ArchonMegalon:


SeanPGorman:
@ArchonMegalon Thanks for all your work on this. Can you give me a basic lesson on this functionality?

ArchonMegalon:
@SeanPGorman well, with the next milestone there should be the first plugin for the plugin-framework present. It's called SINners (Cloud). First you have to enable (experimental) Plugins in the Options. You will than see a new TabPage, where you can select which Plugin you want to enable (the ListBox on the left - don't forget to "check" the Plugin, otherwise it's present but not enabled.

Then you will see the different options to configure my plugin. First you should press the "Login"-button and register with your email address (the whole thing is about being an online repository for users, so there is sadly no way to use it without registering - except maybe search, but that's not implemented yet). 

After registering you should make sure, that your email is added to the "Visibility to" Listbox and the "can edit" checkbox is checked. Otherwise the Chummers you upload wont be visible to you or you wont be able to edit them afterwards. 

Also having the "Visible to my group" would help to join a group and actually see the other members of the group. "Is public visible" is not used for the time being, but I would suggest making you chummers searchable by default. After all, that's what a repository is about.

"Onlinemode" is the most interesting thing right now I think. If you check that, with every click on the save button the character is saved online. If you use chummer on another computer, where you log in with the same user you can open the last version of you chummers by just making a doubleclick on it in the "online" section.

What is currently under development and partly implemented are groups. If you load a chummer, you have a new TabPage at the end of characterTab called "SINners". You can upload you charakter there (if you are logged in) and after that you can create or join a group. You can only use the groupname, Alias and Users are not implemented right now. 

So - upload you character. Don't wonder if it takes a while and you don't get visual feedback for a couple of seconds, because the UI is missing a statusbar or something like that to report the progress of up-/or downloading at the moment. After a while the "search group" button will get enabled. There you can enter a name for a (your) group and search it. If it doesn't exist (only "Test" exists right now), you can press the Create Group-Button and the group will be created and your chummer will join it. If another user than (or you can also add a second chummer to the same group) upload his chummer, searches for your groupname and "joins" the group you will see the character of the other user as a subnode of your character in the Characteroster-Tree when restarting chummer. At least that's the plan ;)

Keep in mind that this is VERY MUCH under development and absolutely NOT stable. :D

BUT please try it out and provide feedback or even better - spread the word about it, so I get some useful data about usage and feedback about what is and is not working or what people are missing on functionality.

Don't wonder about Archetypes, that's for future use and only members of the group "ArchetypeAdmin" can set that flag. Honestly it's more about me and chummer being lazy and we would like some other guys to jump in to administer such things. 👯‍♂️

SeanPGorman:
query: next milestone - is that next nightly or next full version? I just updated to current nightly and cannot see the checkbox for allowing experimental or any plugins.

SeanPGorman:
How do I find a new character @ArchonMegalon

ArchonMegalon:
Search is currently not implemented. It's basically a online archive just for you now like dropbox.
The only "other" character you can download are the members of your group if you joined one. You can join my "Test" group if you want. Open a Character, go to the Sinners-Tab, upload it, click on "search groups", enter "Test" in the name-Textbox and click "search". You should find it an should be able to join it. I THINK you have to restart chummer after that, because I don't yet refresh the characterroster after joining a group. But after the next restart you should see all other groupmembers as subnodes of your char and you should be able to download them. Keep in mind that takes a while and there is no visual feedback yet, so give it (a lot) couple of seconds...

PR #3503, opened 2019-03-18
Ignore complex form limit

MachineMuse:
Hi I added an option to ignore the complex form limit in Career mode cause this is now a houserule on Runnerhub. I don't know how to add it to the houserule adjustment package or how to add it to the language files but I tested the option with and without and it doesn't crash or anything so yeah!

### DIFF
--- a/Chummer/Classes/CharacterOptions.cs
+++ b/Chummer/Classes/CharacterOptions.cs
@@ -1,4 +1,4 @@
-/*  This file is part of Chummer5a.
+﻿/*  This file is part of Chummer5a.
  *
  *  Chummer5a is free software: you can redistribute it and/or modify
  *  it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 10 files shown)


MachineMuse:
I figured out the language files; I dunno how to add it to the houserules package if that's even possible but it should be good to go now.

I only speak English and French well enough to do localizations, hopefully it's not too terrible

chummer5a:
Currently no way to add to a customdata package, but I'd assume you guys would have like a Runnerhub Settings file that has any other arbitrary changes that are made?

MachineMuse:
I felt like it's a house rule that some home games might want as well if they're getting to the point where their technomancer is capped on CFs, so I figured it should be an option.

As for the runnerhub settings file, that's probably a good idea! I know there is a houserules package for Runnerhub already, I've heard it's a bit out of date though. I haven't had time/energy to look into it

PR #3539, opened 2019-04-01
German translation for easter eggs

Ramesch:
I don't know if chummer5 use translation for easter eggs. If so I put it into German language files. :)


chummer5a:
Champ, I'll get the translation management for it integrated at some point.

PR #3612, opened 2019-04-25
Fixed SINners

ArchonMegalon:
SINners should be working again.
Got rid of the Dictionary in frmCharacterRoster. Everything seems to work except sorting (I threw that out for now).
Double uploads are now shown - that should help make clear, why sometimes a character can't be uploaded.


Desertcatkwh:
I will load up the nightly and tonight or tomorrow and edit up Zamriel...-Desertcatkwh

ArchonMegalon:
@Desertcatkwh yes, please give me feedback if it worked. Maybe also some other input? What would you like to have implemented as well (and how)?

Desertcatkwh:
I had to completely disable the SINner plugin but my character is saving again. Thanks.

I will sit and think through further input. Chummer really is good software and want to make sure any comments are helpful and/or insightful.
-Desercatkwh

ArchonMegalon:
If you want to just use Chummer, you should seriously left it disabled for now. SINners is under hefty development and very unstable atm. On the other hand I of course welcome everyone, who is willing to be an alpha tester and doesn't mind putting on with all the crappy bugs that are in there right now - and all of them I introduce with each patch. :D

PR #3654, opened 2019-05-15
NLog Implementation

ArchonMegalon:
Hi @chummer5a !

I have implemented NLog. This is a big change, although it's a pure replacement right now with no additional benefit right of the box (apart from everything the NLog library provides).

Please take a look at it if it still works (it does for me). The NLog.dll itself throws some FirstChanceExceptions in each Log-Call, that's why it's not possible any more to "Log" FirstChanceExceptions like we did previously in the Program startup. 

The Windows Eventlog can only be written to, if Chummer is started once as Admin because NLog needs to register an Eventsource to do that (just in case you take a look at the nlog.config file and wonder, why you can't find entries in your event log).

I replaced all the old "Log.Exception" calls with - what I think - a better "Log.Error(Exception)" or in some cases Log.Trace(Exception), when the exception acutally doesn't matter.

If you like this framework I would implement a new Option like "provide my Logs to the Devs online" and implement a Application Insights target for NLog.

### DIFF
--- a/Chummer.Tests/app.config
+++ b/Chummer.Tests/app.config
@@ -20,6 +20,10 @@
         <assemblyIdentity name="System.Composition.Hosting" publicKeyToken="b03f5f7f11d50a3a" culture="neutral" />
         <bindingRedirect oldVersion="0.0.0.0-1.0.33.0" newVersion="1.0.33.0" />
       </dependentAssembly>
+      <dependentAssembly>
+        <assemblyIdentity name="System.Diagnostics.DiagnosticSource" publicKeyToken="cc7b13ffcd2ddd51" culture="neutral" />
+        <bindingRedirect oldVersion="0.0.0.0-4.0.3.1" newVersion="4.0.3.1" />
+      </dependentAssembly>
     </assemblyBinding>
   </runtime>
 </configuration>
--- a/Chummer/ApplicationInsights.config
+++ b/Chummer/ApplicationInsights.config
@@ -1,0 +1,111 @@
+<?xml version="1.0" encoding="utf-8"?>
+<ApplicationInsights xmlns="http://schemas.microsoft.com/ApplicationInsights/2013/Settings">
+    <InstrumentationKey>012fd080-80dc-4c10-97df-4f2cf8c805d5</InstrumentationKey>
+    <TelemetryInitializers>
+        <Add Type="Microsoft.ApplicationInsights.DependencyCollector.HttpDependenciesParsingTelemetryInitializer, Microsoft.AI.DependencyCollector"/>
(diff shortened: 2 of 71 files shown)


chummer5a:
I'd prefer not to use event log if admin rights are required. I'll have a poke at it later. Happy to expand the logging options, might be worth having it intercept the crash handler as well but I can look into that myself if it's a bit spooky.

ArchonMegalon:
Ok, maybe that was misleading. Admin rights are NOT required at all for Logging to all kinds of targets (like into a file, a database, a webservice, a mailreceiver)

Only if you want to use the EventLog, they are required ONCE, because the Windows-OS prevents none-admins from ... well logging to the EventLog. 

It is perfectly usable even IF the Eventlog-Target is defined and it doesn't have the privileges. It just won't log there (because it's forbidden by the OS to do so).

ArchonMegalon:
Actually - this is not ready yet. Wait a bit... Still fiddling around with application insights...

ArchonMegalon:
Ok, works again! @chummer5a - give it a try. If you tell me your microsoft account I can share the azure-portal for the logging with you.

chummer5a:
Looks fine to me at first blush, couple of things; 

- There's inconsistent use of the logger class being static in forms and equipment classes. Likely minor, but might be relevant for performance?  
- Is it possible to have the applicationinsights export to a file that users can send to us, or is it purely an API push? Super unimportant, but might help us to control the use of it.

Other than that, looks good!

ArchonMegalon:
Well, i also wondered about the sometimes static and sometimes non static use. The unofficial answer I got was, that the GetCurrentClass Function should return a static anyway and it doesn't matter currently.
So we could make it static as well.  There was a BUT afterwards: There may be some changes in the future, that would make use of none-static logger instances, and if we force it to be static, we would not be able to use such features. He was talking about callbacks or such things I think, that are able to register on the logger. 

As far as I know it's not possible to have application insights generate a file. The API for example supports a live stream as well, that would not work that way. Of coures we are logging the same calls as NLog does, so if the user doesnt enable A(pplication) I(nsights), but only "Logging", he can just send us the file as well. AI is really just more comfortable for the end user.

Have you registered a microsoft account already to get access to the AI-Logs?

chummer5a:
Toss it to chummer5isalive@gmail.com

ArchonMegalon:
Can't, you can register here: https://account.microsoft.com

Doku:
To provide access to another user

You must have Owner rights to the subscription or the resource group.

The user must have a **Microsoft Account**, or access to their organizational Microsoft Account. You can provide access to individuals, and also to user groups defined in Azure Active Directory.

PR #3661, opened 2019-05-17
Tried to add metric, but it seems I can't get Filenames/Linenumbers

ArchonMegalon:
@chummer5a (or whoever wrote the heatmap), please take a look at it, because in my debugger there is never a content for "file".

Not sure what's wrong there, but the Metric thing should be nice for a heatmap in ApplicationInsights. 
Also loading a character could be an "operiation" for AI, with measurement how long it takes and what errors it throws (would implement that probably on monday)


chummer5a:
Figured it out or find a better way to handle it? I'm not sure that the heatmap ever got much love when it was first implemented.

ArchonMegalon:
I didn't solve this and decided to not bother anymore. So much else to do. ;)
But if you think this is something interesting we can try to get it to work.

PR #3667, opened 2019-05-21
AI performance log

ArchonMegalon:
It still works (at least for me). Please check that.


ArchonMegalon:
@chummer5a It's working now. We can now see how long the segments take and they are structured hierarchically. I think we can merge this version now. 
 
image

For example I learned that this call is the one slowing the loading into frmCareer down the most:
  tabSkillsUc.RealLoad();
Maybe we can extract it and load this tabPage seperatly after all others are loaded and shown?
See the Chart:

image

Issue #3562, opened 2019-04-09
The Ghosts are not correctly ordered for the Tradition Straßenhexen


azania666:
**Is your feature request related to a problem? Please describe.**
The Ghosts are not correctly ordered for the Tradition Straßenhexen

It is 
Combat: Guidance Spirit
Detection: Spirit of Water
Health: Spirit of Fire
Illusion: Spirit of Air
Manipulation: Spirit of Man

**Describe the solution you'd like**

It should be:
Combat: Spirit of Fire
Detection: Spirit of Man
Health: Guidance Spirit
Illusion: Spirit of Water
Manipulation: Spirit of Air
![Straßengrimoire](https://user-images.githubusercontent.com/43959752/55786695-10444200-5ab5-11e9-947f-ffd6a98c5325.JPG)
![Chummer](https://user-images.githubusercontent.com/43959752/55786697-10444200-5ab5-11e9-9e47-a094938231ad.JPG)

**Describe alternatives you've considered**

**Additional context**
The ghosts listet are in the order printed in the book, but the list of categories is in another Order.

PR #3678, opened 2019-05-30
German translation

Ramesch:
Added German translation for missing strings, homunculi and watcher (also added missing guid for watcher in traditions.xml)


chummer5a:
While you're at it, someone raised an issue a while back about one of the German traditions not having the right spirits, could you take a look and confirm what they need to be? #3562

Ramesch:
Azania666 is right. I put the changes into https://github.com/chummer5a/chummer5a/pull/3680.

PR #3734, opened 2019-07-02
possible fix for Cannot change ObservableCollection during a Collecti…

ArchonMegalon:
Had this crash report. I'm not sure what the add quality has to do with the level of the quality, so could you please review that fix? It basically just makes the original code work, BUT I think the original intention is actually not the right way to do it....

[chummer_crash_132064865460620359.txt](https://github.com/chummer5a/chummer5a/files/3348422/chummer_crash_132064865460620359.txt)

### DIFF
--- a/Chummer/Character Creation/frmCreate.cs
+++ b/Chummer/Character Creation/frmCreate.cs
@@ -7336,7 +7336,17 @@ namespace Chummer
                     if (blnAddItem)
                     {
                         blnRequireUpdate = true;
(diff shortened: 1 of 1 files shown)


chummer5a:
Quality levels were implemented to handle some arcane legacy mechanisms, largely relating to how we remove qualities. I think the theory at the time was that adding them in as separate objects that could then be collapsed would allow for handling existing characters since the qualities could be collapsed down to one instance. Anyway, this looks fine for our purposes.

PR #3744, opened 2019-07-08
fixing things

ArchonMegalon:


Lochabar:
So how am I supposed to sort the character roster?

ArchonMegalon:
Best would be to make a feature request and tag and SCREAM at me, so that i feel guilty enough to re-implement it again. That should work... :)

PR #3763, opened 2019-07-15
Change ApplicationInsights-Setting in GlobalOptions from Bool to Enum…

ArchonMegalon:
… and added translation (en, de)

### DIFF
--- a/Chummer/Backend/Debugging/CrashHandler.cs
+++ b/Chummer/Backend/Debugging/CrashHandler.cs
@@ -192,20 +192,16 @@ namespace Chummer.Backend
                 byte[] info = new UTF8Encoding(true).GetBytes(dump.SerializeBase64());
                 File.WriteAllBytes(Path.Combine(Utils.GetStartupPath, "json.txt"), info);
 
-                if (GlobalOptions.UseLoggingApplicationInsights == EnumUseLoggingApplicationInsights.yes
-                    || GlobalOptions.UseLoggingApplicationInsights == EnumUseLoggingApplicationInsights.crashes)
+                if (GlobalOptions.UseLoggingApplicationInsights >= UseAILogging.crashes)
                 {
                     if (Program.TelemetryClient != null)
                     {
-                        var oldvalue = GlobalOptions.UseLoggingApplicationInsights;
-                        GlobalOptions.UseLoggingApplicationInsights = EnumUseLoggingApplicationInsights.yes;
                         ExceptionTelemetry et = new ExceptionTelemetry(ex)
                         {
                             SeverityLevel = SeverityLevel.Critical
                         };
                         Program.TelemetryClient.TrackException(et);
                         Program.TelemetryClient.Flush();
-                        GlobalOptions.UseLoggingApplicationInsights = oldvalue;
                     }
                 }
 
--- a/Chummer/Backend/Debugging/CrashHandler.cs
+++ b/Chummer/Backend/Debugging/CrashHandler.cs
@@ -192,16 +192,20 @@ namespace Chummer.Backend
                 byte[] info = new UTF8Encoding(true).GetBytes(dump.SerializeBase64());
                 File.WriteAllBytes(Path.Combine(Utils.GetStartupPath, "json.txt"), info);
 
-                if (GlobalOptions.UseLoggingApplicationInsights)
+                if (GlobalOptions.UseLoggingApplicationInsights == EnumUseLoggingApplicationInsights.yes
+                    || GlobalOptions.UseLoggingApplicationInsights == EnumUseLoggingApplicationInsights.crashes)
                 {
                     if (Program.TelemetryClient != null)
                     {
+                        var oldvalue = GlobalOptions.UseLoggingApplicationInsights;
+                        GlobalOptions.UseLoggingApplicationInsights = EnumUseLoggingApplicationInsights.yes;
                         ExceptionTelemetry et = new ExceptionTelemetry(ex)
                         {
                             SeverityLevel = SeverityLevel.Critical
                         };
                         Program.TelemetryClient.TrackException(et);
                         Program.TelemetryClient.Flush();
+                        GlobalOptions.UseLoggingApplicationInsights = oldvalue;
                     }
                 }
 
--- a/Chummer/Backend/Debugging/CrashHandler.cs
+++ b/Chummer/Backend/Debugging/CrashHandler.cs
@@ -192,7 +192,7 @@ namespace Chummer.Backend
                 byte[] info = new UTF8Encoding(true).GetBytes(dump.SerializeBase64());
                 File.WriteAllBytes(Path.Combine(Utils.GetStartupPath, "json.txt"), info);
 
-                if (GlobalOptions.UseLoggingApplicationInsights >= UseAILogging.crashes)
+                if (GlobalOptions.UseLoggingApplicationInsights >= UseAILogging.Crashes)
                 {
                     if (Program.TelemetryClient != null)
                     {
--- a/Chummer/Classes/clsOptions.cs
+++ b/Chummer/Classes/clsOptions.cs
@@ -232,7 +232,7 @@ namespace Chummer
         private static string _strPDFParameters = string.Empty;
         private static HashSet<SourcebookInfo> _lstSourcebookInfo;
         private static bool _blnUseLogging;
-        private static int _blnUseLoggingApplicationInsights;
+        private static EnumUseLoggingApplicationInsights _enumUseLoggingApplicationInsights;
         private static string _strCharacterRosterPath;
 
         // Custom Data Directory information.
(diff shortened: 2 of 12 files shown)


ArchonMegalon:
Hi! If changed the option to tweak how Application Insights is used. Please take a look at it, since this is somewhat sensible. I like the Enum-Idea from joha and implemented it.

Chummer/Backend/Debugging/CrashHandler.cs
```diff
@@ -192,16 +192,20 @@ public static void WebMiniDumpHandler(Exception ex)
...
                 File.WriteAllBytes(Path.Combine(Utils.GetStartupPath, "json.txt"), info);
 
-                if (GlobalOptions.UseLoggingApplicationInsights)
+                if (GlobalOptions.UseLoggingApplicationInsights == EnumUseLoggingApplicationInsights.yes
+                    || GlobalOptions.UseLoggingApplicationInsights == EnumUseLoggingApplicationInsights.crashes)
                 {
                     if (Program.TelemetryClient != null)
                     {
+                        var oldvalue = GlobalOptions.UseLoggingApplicationInsights;
+                        GlobalOptions.UseLoggingApplicationInsights = EnumUseLoggingApplicationInsights.yes;
```
joha4270: This is _not_ the correct way of enabling telemetry for crashes. 
Properties are supposed to be fancy variables, this is specifically relying on side effects, of writing to a property, without even a comment. 

I would simply write to the telemetry variable, if this is not a good solution for some reason, create a TelemetrySomething that both Options and this can interact with.

Chummer/Classes/clsOptions.cs
```diff
@@ -45,6 +45,15 @@ public enum ClipboardContentType
         Lifestyle,
     }
 
+    public enum EnumUseLoggingApplicationInsights
```
joha4270: We got intellisense, we don't need the name to also tell us its an enum. There is a lot of old code in chummer that does this, and a lot of old programmers that does this, but unless you know something than the typechecker don't, it is only extra text I have to read.

Chummer/Classes/clsOptions.cs
```diff
@@ -399,7 +408,9 @@ static GlobalOptions()
             LoadBoolFromRegistry(ref _blnUseLogging, "uselogging");
 
             //Should the App "Phone home"
-            LoadBoolFromRegistry(ref _blnUseLoggingApplicationInsights, "useloggingApplicationInsights");
+            string myuseAI = "notset";
```
joha4270: _Consider_ (no more, this is a matter of debate) enclosing those 3 lines in a block `{}`. This will prevent `myuseAI` poluting the outer scope, at the cost of some extra indentation. 

Also, _my_useAI, I don't think the my really adds any information to the variable name, but I'm nitpicking here

Chummer/Classes/clsOptions.cs
```diff
@@ -622,16 +633,16 @@ public static bool UseLogging
         /// <summary>
         /// Whether or not the app should use logging.
         /// </summary>
-        public static bool UseLoggingApplicationInsights
+        public static EnumUseLoggingApplicationInsights UseLoggingApplicationInsights
         {
-            get => _blnUseLoggingApplicationInsights;
+            get => (EnumUseLoggingApplicationInsights)Enum.Parse(typeof(EnumUseLoggingApplicationInsights), _enumUseLoggingApplicationInsights.ToString());
```
joha4270: Are you converting the enum to a string, parsing it again, then casting it to its original type?
Is there a reason `get => _enumUseLoggingApplicationInsights;` does not work?

Chummer/Classes/clsOptions.cs
```diff
@@ -622,16 +633,16 @@ public static bool UseLogging
...
             set
             {
-                if (_blnUseLoggingApplicationInsights != value)
+                if (_enumUseLoggingApplicationInsights != value)
                 {
-                    _blnUseLoggingApplicationInsights = value;
+                    _enumUseLoggingApplicationInsights = value;
                     // Sets up logging if the option is changed during runtime
-                    if (value)
+                    if ((int)_enumUseLoggingApplicationInsights < 1)
```
joha4270: You can compare enums directly, so `if (_enumUseLoggingApplicationInsights < EnumUseLoggingApplicationInsights.onlymetric)` has same effect without any magic values of 1

Furthermore your if can be entirely removed and simplified to `TelemetryConfiguration.Active.DisableTelemetry = !(_enumUseLoggingApplicationInsights < EnumUseLoggingApplicationInsights.onlymetric)`
(Which, if I'm parsing it correctly, enables logging if `_enumUseLoggingApplicationInsights` one of `notset`, `onlylocal`, `onlymetric`, which I don't think is the intended behavior)

Chummer/Program.cs
```diff
@@ -150,10 +150,14 @@ static void Main()
                 try
                 {
                     LogManager.ThrowExceptions = true;
-                    ConfigurationItemFactory.Default.Targets.RegisterDefinition(
-                        "ApplicationInsightsTarget",
-                        typeof(Microsoft.ApplicationInsights.NLogTarget.ApplicationInsightsTarget)
-                    );
+                    if (GlobalOptions.UseLoggingApplicationInsights != EnumUseLoggingApplicationInsights.onlylocal)
```
joha4270: I don't think this is the intended operator, did you intend `>` (`>=EnumUseLoggingApplicationInsights.onlymetric`)?

Chummer/Program.cs
```diff
@@ -167,10 +171,11 @@ static void Main()
                     }
                     Log.Info(strInfo);
 
-                   
 
-                    if (GlobalOptions.UseLoggingApplicationInsights)
+                    if (GlobalOptions.UseLoggingApplicationInsights != EnumUseLoggingApplicationInsights.onlylocal
+                        && GlobalOptions.UseLoggingApplicationInsights != EnumUseLoggingApplicationInsights.notset)
```
joha4270: Always true. (if not equal to 2 different values at the same time)

Chummer/frmOptions.cs
```diff
@@ -1867,11 +1900,13 @@ private void clbPlugins_ItemCheck(object sender, ItemCheckEventArgs e)
...
         }
 
-        private void chkUseLoggingApplicationInsights_CheckedChanged(object sender, EventArgs e)
+        private void cbUseLoggingApplicationInsights_SelectedIndexChanged(object sender, EventArgs e)
         {
             if (this._blnLoading)
                 return;
-            if (chkUseLoggingApplicationInsights.Checked)
+            EnumUseLoggingApplicationInsights useAI =
+                (EnumUseLoggingApplicationInsights)Enum.Parse(typeof(EnumUseLoggingApplicationInsights), ((ListItem)cbUseLoggingApplicationInsights.SelectedItem).Value.ToString());
```
joha4270: I think selectedItem should have the type `EnumUseLoggingApplicationInsights` already and you can just cast it?

Chummer/Program.cs
```diff
@@ -248,15 +255,10 @@ static void Main()
                 MainForm = new frmChummerMain(false, pvt);
                 Application.Run(MainForm);
                 Log.Info(ExceptionHeatmap.GenerateInfo());
-                if (GlobalOptions.UseLoggingApplicationInsights)
+                if (GlobalOptions.UseLoggingApplicationInsights != EnumUseLoggingApplicationInsights.onlylocal)
```
joha4270: Again, you probably want `>` or something similar

Chummer/frmOptions.cs
```diff
@@ -1888,27 +1923,34 @@ private void chkUseLoggingApplicationInsights_CheckedChanged(object sender, Even
...
             else
             {
-                GlobalOptions.UseLoggingApplicationInsights = false;
+                GlobalOptions.UseLoggingApplicationInsights = useAI;
             }
         }
 
         private void ChkUseLogging_CheckedChanged(object sender, EventArgs e)
         {
             if (this.chkUseLogging.Checked)
```
joha4270: No need for an if, just  `.Enabled = .Checked`

Chummer/frmOptions.cs
```diff
@@ -1888,27 +1923,34 @@ private void chkUseLoggingApplicationInsights_CheckedChanged(object sender, Even
...
-                this.chkUseLoggingApplicationInsights.Enabled = true;
+                this.cbUseLoggingApplicationInsights.Enabled = true;
             else
-                this.chkUseLoggingApplicationInsights.Enabled = false;
+                this.cbUseLoggingApplicationInsights.Enabled = false;
             OptionsChanged(sender, e);
         }
+
+        private void chkUseLoggingApplicationInsights_CheckedChanged(object sender, EventArgs e)
```
joha4270: Empty method?

Chummer/Classes/clsOptions.cs
```diff
@@ -45,6 +45,15 @@ public enum ClipboardContentType
         Lifestyle,
     }
 
+    public enum EnumUseLoggingApplicationInsights
+    {
+        notset = -1,
```
joha4270: Also, by convention enum values are written in PascalCase, not lowercase.

Chummer/Backend/Debugging/CrashHandler.cs
ArchonMegalon: That actually was already legacy. I got rid of it.

Chummer/Classes/clsOptions.cs
joha4270: Well, the _actually_ correct solution would be `LoadEnumFromRegistry<T>`, but...

ArchonMegalon:
Interesting how this review thing works here in github. It's my first time I see it in action. I pushed the changes joha requested (thanks for the detailed intructions INCLUDING a description). :)

If it's ok now - can you merge it? anything for me left to do?

PR #3942, opened 2019-11-22
Cyberzombie

biscuitWizard:


chummer5a:
Appreciate the effort, but it'd been bugging me for ages that it was driven by names; I've made it be an improvement type instead.

PR #3943, opened 2019-11-24
Add qualities for Dark Terror advanced optional powers for Nosferatu

Malkleth:
Add qualities to give Nosferatu PCs access to Mimicry and Psycokinesis. I couldn't get a quality for Infected Mystic Armor to work, although I did add a rating tag. It can already be added via the critter powers tab for appropriate karma.


chummer5a:
Yeah, critter powers aren't really built properly to handle having levels. Need to sort something out with that, but since there's none of the restrictions that adept powers have, it's a bit quirky. I'd rather not just set the levels to be unlimited, but that sort of seems like how it's supposed to be.

PR #3945, opened 2019-11-25
Changed Program.Main to launch a ShadowCopy AppDomain. 

ArchonMegalon:
That should help with the locked files problem while updating.


ArchonMegalon:
not ready yet...

PR #3946, opened 2019-11-26
[WIP] Fixes #3894 - Only Use Highest Foci Bonus For Spells

### LINKED ISSUE #3894: Power and combat spellcasting focus both adding to dice pools in 5e
**Describe the bug**
Both Power focus and combat spellcasting focus are adding to combat dice pools

**To Reproduce**
1. Go to Gear
2. Click on Add Gear
     Add power focus rank 1
     Add Spellcasting focus, combat rank 2
     Bind both foci
3. Go to Skills, Spellcasting, see Power focus added to dice pool
4. Go to Spells and Spirits, any combat spell, see Spellcasting Focus, combat added to dice pool

**Expected behavior**
Only the highest force focus should add to the dice pool

**Screenshots**

**Character File**
If applicable, add your character file to help explain your problem. Character files can be compressed into zips and uploaded here, or have their extension renamed to .txt.

 - OS:  Windows 10
 - Version 5.208.0

chummer_spellcasting
chummer_manabolt

MemeticContagion:
Hi! I saw this was open and thought I would take a stab at fixing it. This is the first time I've done anything with .NET, so please let me know if this is the wildly wrong approach. I'll try to comment the PR where I have questions or I think I'm doing anything especially strange. 

This checks on calculating the spell dice poll to see if the character has other bonded foci, and if so, if any of them are power foci that are added to the base MAG score. If there are any power foci with a value higher than the specialized focus, it just skips adding in the specialized focuses value. 

If there is a power focus but the highest value one is below the rating of the specialized focus, it gives a bonus for the specialized focus that is the diff of the power focus - specialized focus. My current approach to doing that is almost certainly non-standard.

### DIFF
--- a/Chummer/Backend/Uniques/Spell.cs
+++ b/Chummer/Backend/Uniques/Spell.cs
@@ -849,12 +849,15 @@ namespace Chummer
                             yield return objImprovement;
                         break;
                     case Improvement.ImprovementType.SpellCategory:
-                        // We need to do some checking to make sure this is the most powerful foci before we add it in
-                        if (objImprovement.ImproveSource == Improvement.ImprovementSource.Gear )
+                        // SR5 318: Regardless of the number of bonded foci you have,
+                        // only one focus may add its Force to a dicepool for any given test.
+                        // We need to do some checking to make sure this is the most powerful focus before we add it in
+                        if (objImprovement.ImproveSource == Improvement.ImprovementSource.Gear)
                         {
+                            //TODO: THIS IS NOT SAFE. While we can mostly assume that Gear that add to SpellCategory are Foci, it's not reliable.
                             // we are returning either the original improvement, null or a newly instantiated improvement
-                            Improvement? bestFocus = CompareFocusPower(objImprovement);
-                            if (bestFocus is Improvement)
+                            Improvement bestFocus = CompareFocusPower(objImprovement);
+                            if (bestFocus != null)
                             {
                                 yield return bestFocus;
                             }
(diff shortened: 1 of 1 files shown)


Chummer/Backend/Uniques/Spell.cs
```diff
@@ -859,6 +874,42 @@ private IEnumerable<Improvement> RelevantImprovements(Func<Improvement, bool> fu
...
+                _objCharacter.Foci.Where(x => x.GearObject.Bonus.InnerText == "MAGRating" && x.GearObject.Bonded);
+            int powerFocusRating = fociList.Any()
+                ? fociList.OrderByDescending(x => x.Rating)?.FirstOrDefault().Rating ?? default(int)
+                : default(int);
+            // If our focus is higher, add in a partial bonus
+            if (powerFocusRating > 0 && powerFocusRating < objImprovement.Value)
+            {
+                // This is hackz -- because we don't want to lose the original improvement's value
+                // we instantiate a fake version of the improvement that isn't saved to represent the diff
+                return new Improvement(_objCharacter)
```
MemeticContagion: This is the weirdest part of this PR, I think--we're basically creating a fake improvement representing the diff. But we don't want to modify the real improvement's value, since if the power focus is ever unbonded, we'll end up wanting to put the full value improvement back into play. Open to suggestions!

Chummer/Backend/Uniques/Spell.cs
```diff
@@ -849,6 +849,21 @@ private IEnumerable<Improvement> RelevantImprovements(Func<Improvement, bool> fu
                             yield return objImprovement;
                         break;
                     case Improvement.ImprovementType.SpellCategory:
+                        // We need to do some checking to make sure this is the most powerful foci before we add it in
+                        if (objImprovement.ImproveSource == Improvement.ImprovementSource.Gear )
```
MemeticContagion: Is this the best way to see if the SpellCategory improvement is a focus? Is there a narrower category I should be using?

Chummer/Backend/Uniques/Spell.cs
```diff
@@ -859,6 +875,42 @@ private IEnumerable<Improvement> RelevantImprovements(Func<Improvement, bool> fu
...
         }
+
+        /// <summary>
+        /// Method to check we are only applying the highest focus to the spell dicepool
+        /// </summary>
+        private Improvement CompareFocusPower(Improvement objImprovement)
+        {
+            // get any bonded foci that add to the base magic stat and return the highest rated one's rating
+            IEnumerable<Focus> fociList =
+                _objCharacter.Foci.Where(x => x.GearObject.Bonus.InnerText == "MAGRating" && x.GearObject.Bonded);
```
MemeticContagion: Is this the best way to find any foci that add directly to the magic rating?

Chummer/Backend/Uniques/Spell.cs
chummer5a: Yeah, there's not really anything cuter than that. You could ghetto something with GetSourceName, but it'd be a kludge.

chummer5a:
I'd have preferred to have seen something data-driven, but this is actually fairly elegant 👍

PR #3961, opened 2019-12-24
Fixed custom rule 'German Data Changes - Static Shaft Crossbow Bolts'

Ramesch:
- renamed 'amend_crossbow_weapons.xml' to 'amend_bolts_gear.xml' as it should add ammunition (gear, not weapon)
- added 'hide'-tag to normal 'Bolt: Stick-n-Shock w/Static Shaft' in custom rule as this ammunition probably should be replaced
- added German translation for bolts in custom rule 'German Data Changes - Static Shaft Crossbow Bolts'


chummer5a:
That's what I get for not testing amend files. Cheers for fixing it properly

PR #3970, opened 2019-12-30
Fixed non-selectable custom rules vor German Data Changes

Ramesch:
- added 'amendoperation'-tag to non-selectable custom rules 'German Data Changes ...'
- verified content and added missing custom rules 'German Data Changes ...' to 'Chummer.csproj'


chummer5a:
Specifying amendoperation="replace" for nodes that exist generally isn't necessary; the amend system will automatically detect whether it should be adding or editing a node. Looks fine other than that though.

PR #3972, opened 2019-12-30
The free grid subscriptions are not added to lifestyles as free quali…

batchii:
…ties unless this house rule is checked or the player has set the Hard Targets book to active.

This is my first PR so please let me know if there is anything that should be changed, or if I missed following any guidlines.


chummer5a:
Looks good! As a general rule we only add translation keys for en-us or whatever languages you do know, as the entry will otherwise be flagged as translated. Not hugely important though. 
Also we use a lowercase standard for all our XML nodes, so if you could also update the 'allowFreeGrids' node to be lowercase it'll be good to merge in.

Ramesch:
German translation for 'Allow Free Grid Subscription Qualities for lifestyles even if Hard Targets is not an active book' would be:
- Zulassen von beliebigen Gitter-Abonnements für Lebensstile, auch wenn 'Harte Ziele' nicht als Quellenbuch ausgewählt ist

So if you want, you can take this string for 'Checkbox_Options_AllowFreeGrids' in 'de-de.xml'.

PR #3975, opened 2019-12-31
Addition to custom rule for German Yamaha Kaburaya

Ramesch:
Besides the acceleration, the costs for the motorbike 'Yamaha Kaburaya' are also different according to German rules. Since this is mentioned not only once but twice and is not mentioned in any errata thread in German forums, this adjustment seems to be intentional. The custom rule has been modified to include both.

Licence text has been added.


Ramesch:
I understand that 'replace' is not necessary in 'amendoperation'. Thanks for the hint.

But I will continue to use it, because it helps me to easily find out what is added, what is changed and what is replaced in the rule files. Is that ok or is that against the conventions?

chummer5a:
Nah, using replace is completely fine. Just unnecessary, but that does no harm.

Issue #3954, opened 2019-12-08
Feature request: dropdown menu for initiation


Harlekin13:
I really struggled to find the possibility to add metamagics to initiated characters, I stumbled across it when accidentialy rightclicking on the grade.
Perhaps it would be good to add a similar dropdown menu to the button "add initiate grade" as it is there on "add gear" or "add cyberware"? That way one can easily see there is more on that tab than only adding initiate grades.

ArchonMegalon:
Closed with #3978

PR #3978, opened 2020-01-02
Made Add Initiation Button a Splitbutton

ArchonMegalon:
See #3954


chummer5a:
Create mode?

ArchonMegalon:
Still missing - did it only in frmCareer right now. Will do in Create as
well...

PR #3987, opened 2020-01-09
Translation

Ramesch:


Ramesch:
@gdelmee Thank you very much. My French is from school days and I only rarely use it to read texts. In Shadowrun's case, it is also a fact that the translations of some words are not literal, but have a different origin in each language. Among other things, I do not have the French rulebooks with names and other typical terms. If you could improve the French translation, that would be great.

PR #4005, opened 2020-01-18
Fr translations

gdelmee:
Moving forward in fr-fr check


chummer5a:
Sulking about conflicting file, prboably just needs to be squashed

gdelmee:
I've moved lots of strings and done corrections. I understand the system has issues to auto merge.
I'm far from the end of the file, currently at line 2047.
If you prefer, I can continue and we can wait untill the file is completely done.

ArchonMegalon:
That's weird. The conflicts seem to not actually be conflicts, since the master just seems to have more lines. Did you REMOVE lines? Why so? Or am I reading that merge wrong?

gdelmee:
There was 9778 lines and I pushed a file of 9780 lines as I replicated regions, and so added to add end region lines.
Also, there are a lot of changed lines as Visual Studio corrected tabs (with tabify) even there was no changes.

gdelmee:
I've double checked and I did made a mistake. I'll correct it for a further pull request.

PR #4151, opened 2020-06-26
Added approved metagenic qualities

JokerRouge:
ShadowNet approved the use of many qualities as positive and negative metagenic qualities for SURGE characters. Changes reflect the addition of the metagenic TRUE tag for these qualities.


aksu560:
ShadowNET Upkeep head here to confirm this change.

PR #4217, opened 2020-09-10
ChummerHub fixes

ArchonMegalon:


ChummerHub.Client/Model/CharacterExtended.cs
```diff
@@ -159,7 +159,7 @@ internal IList<Tag> PopulateTags()
                 MySINnerFile.SiNnerMetaData.Tags.Remove(f);
             }
             MySINnerFile.SiNnerMetaData.Tags.Add(tag);
-            foreach(var childtag in MySINnerFile.SiNnerMetaData.Tags)
+            foreach(var childtag in MySINnerFile.SiNnerMetaData.Tags.Where(a => a != null).ToList())
```
DelnarErsike, line 162: Just curious, why is the `ToList()` necessary? Can `SetSinnerIdRecursive` add or remove elements from `MySINnerFile.SiNnerMetaData.Tags`?

ArchonMegalon:
No, but since changing to .net core 3.x I got a

"There is already an open DataReader associated with this Connection which
must be closed first." Error, when I cycle through the tags and call
another async operation in the loop. So - since I don't want to rewrite the
inner function (and also expose it as API), i decided to take the
performance hit and just cast it to List here.

MultipeActiveResultSets seems not to be an option for azure-dbs. At least
not in the price-tier we are operating.

Issue #4229, opened 2020-09-30
Cannot Save to SINner directory using "onlinemode"


ghost2501:
While in "onlinemode" using the latest Nighty, I attempt to save an edit to a character that has been previously uploaded but I get the error window (twice) showing the error below:

-----------------------
frmSINnerResponse

Response: Object reference not set to an instance of an object.

Details: ChummerHub.API.HubException: Object reference not set to an instance of an object. ---> System.NullReferenceException: Object reference not set to an instance of an object.
   at ChummerHub.Controllers.V1.SINnerController.PostSINnerInternal(UploadInfoObject uploadInfo) in /src/ChummerHub/Controllers/V1/SINnerController.cs:line 770
   --- End of inner exception stack trace ---

Installation Id: eb7547e4-6a60-495f-9916-347cea8529b6

-----------------------

If I close the character after this, I get the notice that there are unsaved changes. I click yes to save (again) and get this error:

-----------------------
Response: This is NOT an exception from Chummer itself, but from the SINners WebService. This error happend "in the cloud": Parameter is not valid.

Details: System.ArgumentException: Parameter is not valid.
   at System.Drawing.Image.Save(Stream stream, ImageCodecInfo encoder, EncoderParameters encoderParams)
   at Chummer.ImageExtensions.ToBase64String(Image imgToConvert, ImageCodecInfo objCodecInfo, EncoderParameters lstEncoderParameters) in C:\projects\chummer\Chummer\Backend\Extensions\ImageExtensions.cs:line 117
   at Chummer.ImageExtensions.ToBase64StringAsJpeg(Image imgToConvert, Int32 intQuality) in C:\projects\chummer\Chummer\Backend\Extensions\ImageExtensions.cs:line 138
   at Chummer.GlobalOptions.ImageToBase64StringForStorage(Image objImageToSave) in C:\projects\chummer\Chummer\Classes\clsOptions.cs:line 1077
   at Chummer.Character.SaveMugshots(XmlTextWriter objWriter) in C:\projects\chummer\Chummer\Classes\clsCharacter.cs:line 6556
   at Chummer.Character.Save(String strFileName, Boolean addToMRU, Boolean callOnSaveCallBack) in C:\projects\chummer\Chummer\Classes\clsCharacter.cs:line 1210
[...]

PR #4230, opened 2020-09-30
Bugfixes for ChummerHub using docker

ArchonMegalon:


ArchonMegalon:
Hi Delnar! Since you occasionally reviewed my code and I only committed this because of #4229 I would appreciate if you could take a look at this as well if you like. 

The thing I would like to add for the release version would be an azure vault connection to get rid of the OAuth2 secrets in code. The normal configuration does work for normal web apps in azure, but NOT for docker hosted ones (and I have not found anything in the documentation why it woudln't

ChummerHub/Services/GoogleDrive/DriveHandler.cs
```diff
@@ -76,27 +76,30 @@ private UserCredential AuthorizeGoogleUser()
...
-
+                var foolGitGuardian = "-";
                 var flow2 = new GoogleAuthorizationCodeFlow(new GoogleAuthorizationCodeFlow.Initializer
                 {
                     ClientSecrets = new ClientSecrets
                     {
-                        ClientId = Configuration["Authentication.Google.GoogleChummerSINersId"],
-                        ClientSecret = Configuration["Authentication.Google.GoogleChummerSINersSecret"]
+                        ClientId = "779360551859-i817g72s0ork3bffvnhtvpl0q2gi8sub.apps.googleusercontent.com",//Startup.AppSettings["Authentication.Google.GoogleChummerSINersId"],
+                        ClientSecret = "Q2yMsXBtdd" + foolGitGuardian + "zxp6vPXcjkGFz"//Startup.AppSettings["Authentication.Google.GoogleChummerSINersSecret"]
```
ArchonMegalon, line 98: This here is what should NOT happen, I know. BUT both Configuration and Startup.AppSettings are empty in docker...
It works in the WebApp-Slot hosted as a WebApp...

DelnarErsike: Are you not worried about security implications? If you have a single Oauth token set for all users, then in theory, any one of them could mess with data that isn't supposed to be "theirs", not to mention possible issues from just writing out the token in plaintext in a public Git repository for all to see.

ChummerHub/Services/GoogleDrive/DriveHandler.cs
```diff
@@ -109,21 +112,17 @@ private UserCredential AuthorizeGoogleUser()
...
-#pragma warning restore CS1591 // Missing XML comment for publicly visible type or member 'DriveHandler.DriveHandler(ILogger<Startup>, IConfiguration)'
         {
             Configuration = configuration;
             _logger = Logger;
-            string refreshToken = Configuration["Authentication.Google.RefreshToken"];
+            string refreshToken = Startup.AppSettings["AuthenticationGoogleRefreshToken"];
 
             if (string.IsNullOrEmpty(_folderId))
-                _folderId = Configuration["Authentication.Google.ChummerFolderId"];
+                _folderId = "1VHXJk1IbVj0aRQHzuM3-j_fPsc-hCWO7";//Startup.AppSettings["Authentication.Google.ChummerFolderId"];
```
DelnarErsike, line 125: I assume this is a similar situation as in the previous case with Ids being set explicitly instead of from the configuration in docker.

DelnarErsike (commented):
Assuming you're fine with the possible security issues of your Oauth setup, looks OK aside from the null getter thing that I raised in a comment in ChummerHub/Controllers/V1/SINnerController.cs. Once that is handled in some way (potentially even by just saying that yes, that code is intentional), I'll explicitly approve.

ChummerHub/Controllers/V1/SINnerController.cs
```diff
@@ -860,8 +862,8 @@ private async Task<ActionResult<ResultSinnerPostSIN>> PostSINnerInternal(UploadI
                                 user, _context,
                                 _logger, oldgroup.PasswordHash, roles, tc);
                         }
-
-                        if (oldsinner == null)
+                        await _context.SaveChangesAsync();
+                        if (oldsinner == null && user.FavoriteGroups != null)
```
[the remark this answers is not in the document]
ArchonMegalon: Hi! Yes, I got the NullReferenceException. "Problem" is that it was loaded from the database that way I think. In the beta-database there are sometimes messed up entities, that are relics of debugging sessions (like - only half saved) or plain old bugs. Of course I should clean up the database, but.... i thought I would like to make the code as resilient as possible.

ChummerHub/Services/GoogleDrive/DriveHandler.cs
```diff
@@ -76,27 +76,30 @@ private UserCredential AuthorizeGoogleUser()
...
             try
             {
-                string refreshToken = Configuration["Authentication.Google.RefreshToken"];
+                string refreshToken = Startup.AppSettings["AuthenticationGoogleRefreshToken"];
 
                 if (string.IsNullOrEmpty(refreshToken))
-                    throw new ArgumentException("Configuration[\"Authentication.Google.RefreshToken\"] == null! ");
+                    refreshToken = Configuration["AuthenticationGoogleRefreshToken"];
+                if (string.IsNullOrEmpty(refreshToken))
+                    refreshToken = "1/-zsfciq55d9xfAYQ_-U1tmpsMiwHT7oKf1fEO8bm9hQ";
```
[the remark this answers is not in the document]
ArchonMegalon: Well, acutally I would like to figure out how to use Azure Keyvault to get rid of the token on github for good. But that seems not to be that straight forward and since I sadly don't have much time right now (and really dont care that much about the OAuth token), for now I think it's fine here...

ChummerHub/Services/GoogleDrive/DriveHandler.cs
ArchonMegalon: Yes, that is exactly it. On the other hand - worst case is, he deletes all files - since this right now is still early alpha and only temporary...

ChummerHub/Services/GoogleDrive/DriveHandler.cs
ArchonMegalon: yes. It's a pita. FolderId and Refreshtoken are not so senisible, just the Secret is kind of worrieing...

PR #4268, opened 2020-12-06
1.0 Version of HomeTurfNetwork Patreon Customdata Plugin for Chummer5a

Xenomorph-Alpha:
Hello,
created a customdata plugin for my patreon SR5 content. If you have questions you can reach me here or on your discord server.

### DIFF
--- a/Chummer/customdata/HomeTurfNetwork - Patreon Stuff/custom_books.xml
+++ b/Chummer/customdata/HomeTurfNetwork - Patreon Stuff/custom_books.xml
@@ -1,0 +1,28 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--This file is part of Chummer5a.
+
+    Chummer5a is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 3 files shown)


ArchonMegalon:
So I'm unsure of what this would be. Let's keep the discussion here (and not on discord). These are just amend files and it is not a plugin. Do you want us to publish them with the base chummer installation? Do you want it to be a plugin in chummer, that downloads these only for registered patreon users?

What is the "endgame" of this?

Xenomorph-Alpha:
Hi ArchonMegalon,

so the patreon is about doing Shadowrun stuff BUT! everything is free. We can not make people pay for SR stuff because CGL licensing etc. so everything shadowrun related is free. For Patreons we have roleplay maps, fancy versions of the PDFs wich images and our customdata files for chummer as prerelease which will get updated everytime we add something. So we will distribute it ourselves we think that it would be great for people who do not know the patreon they can use it to if they want.


I though such customdata files where declared as plugin, i will look into it. But the Custom Files are ok, because we will just add thing a not change the normal Shadowrun.

PR #4271, opened 2020-12-09
Added OSVersion Metric and app.manifest

ArchonMegalon:


ArchonMegalon:
Hi! I wondered, why you didn't put the app.manifest in the main folder. Can you check? I basically only copied your file...

Chummer/Chummer.csproj
```diff
@@ -91,7 +91,7 @@
   </PropertyGroup>
   <PropertyGroup />
   <PropertyGroup>
-    <ApplicationManifest>Properties\app.manifest</ApplicationManifest>
```
DelnarErsike, line 94: Uh, pretty sure we already have an app.manifest file, and it's in the location that's referred to here. Is there something with its location?

Chummer/helpers/Application Insights/OSVersion.cs
```diff
@@ -0,0 +1,106 @@
...
+        {
+            try
+            {
+                //Get Operating system information.
+                OperatingSystem os = Environment.OSVersion;
+                //Get version information about the os.
+                Version vs = os.Version;
+
+                //Variable to hold our return value
+                string operatingSystem = "";
```
DelnarErsike, line 24: `string.Empty` literal might be cleaner.

Chummer/helpers/Application Insights/OSVersion.cs
```diff
@@ -0,0 +1,106 @@
...
+                            operatingSystem = "10";
+                            break;
+                        default:
+                            break;
+                    }
+                }
+                //Make sure we actually got something in our OS check
+                //We don't want to just return " Service Pack 2" or " 32-bit"
+                //That information is useless without the OS version.
+                if (operatingSystem != "")
```
DelnarErsike, line 83: `!string.IsNullOrEmpty()` might be cleaner.

Chummer/helpers/Application Insights/OSVersion.cs
```diff
@@ -0,0 +1,106 @@
...
+                {
+                    //Got something.  Let's prepend "Windows" and get more info.
+                    operatingSystem = "Windows " + operatingSystem;
+                    //See if there's a service pack installed.
+                    if (os.ServicePack != "")
```
DelnarErsike, line 88: `!string.IsNullOrEmpty()` might be cleaner.

DelnarErsike (commented):
We already have an `app.manifest`, is there something wrong with the way it loads?

Chummer/Chummer.csproj
ArchonMegalon: Yes, we do. I just moved it to the main folder. I wondered if it's actually loaded, because I get get so many Windows 8 installations and the MS-Docs say, that the Environment.OSVersion returns Windows 8 if Windows 10 is not specifically supported in the app.manifest file. Do you mind moving the file?

DelnarErsike (approved):

PR #4281, opened 2020-12-28
Update README.md

chummer5a:


Lochabar:
The introductory paragraph ends with "supports four languages: English, French, German, and Japanese." - it's missing some. It might be better to say "supports multiple languages: English (US), French, German, Japanese, Portuguese (Brazil) and Chinese." (Note: zh-cn is listed on https://www.andiamo.co.uk/resources/iso-language-codes/ as "Chinese (PRC)" if you want to be strict about the version.)

Issue #3861, opened 2019-09-12
Thrown weapons ammunition is always "1"


Lochabar:
**Describe the bug**
Thrown weapons listed in the Ranged Weapons block are always listed as "1" even if there are more that 1 in the gear block.

**To Reproduce**
1. Using a character with more than 1 in a particular thrown weapon
2. Print the character
3. Scroll down to 'Ranged Weapons'
4. See the "Ammo" column for all thrown weapons always contains 1
5. Scroll down to 'Gear'
6. See the Qty for the thrown weapon is more than 1

**Expected behavior**
The ammo column for thrown weapons show the quantity of the thrown weapon.

**Screenshots**
Entries from Gear
![gear](https://user-images.githubusercontent.com/17562417/64749066-dd75fe00-d4d9-11e9-81eb-5008821fec60.png)

Corresponding entries in Thrown Weapons
![ranged](https://user-images.githubusercontent.com/17562417/64749086-f088ce00-d4d9-11e9-91aa-d1d7a9cf41df.png)

**Desktop (please complete the following information):**
 - OS: Windows 7 Home
 - Version: 5.201.12 [and earlier]

**Additional context**
The saved XML file has an ammo node under the thrown weapon of "1". The corresponding gear node has qty of the correct value.

chummer5a:
Hrm. That'd be because as far as the weapon is concerned, it has a clip size of 1 and doesn't know about loose ammo. Kind of a pain to refactor it to work differently for just those. Easiest alternative would be making an 'available ammo' int.

Lochabar:
Requested changes applied with #4308

Issue #2999, opened 2018-09-16
Show notes for skills in Character Viewer / printed character sheet


ghost:
**Is your feature request related to a problem? Please describe.**
You can add notes to skills (as to many other items in Chummer). But these notes are not shown in Character Viewer / printed character sheets.

**Describe the solution you'd like**
As information of eg 'Material Art' for 'Unarmed Combat' is shown maybe notes to skills could also be added to Character Viewer / character sheet. So eg information like 'add 3 dices for visual' could be shown in character sheet.

**Additional context**
Maybe layout is affected too much when adding notes to skills are printed?

PR #4308, opened 2021-02-09
Additions to Notes character sheet

Lochabar:
**Notes character sheet**
Issue #2999 - Notes for skills be shown in the Character Viewer
- added them to the Notes Character Sheet to avoid excessive layout changes in the Shadow 5* sheets.

1) There are a number of common xslt files between used in the Notes and Shadowrun 5* sheets, so many of the css definitions used by the Notes sheet have to duplicate those in the Shadowrun 5* sheets. 
Therefore replaced the css definitions in the "Notes set.xslt" with the the "Shadowrun 5 set CSS.xslt" to keep the definitions in sync.
Note: the base Notes.xsl file and the language specific files have all be modified to import the "Shadowrun 5 set CSS.xslt" file.

2) Reviewed character file to identify as many notes nodes as possible.

Added sections to the Notes sheet to display these notes:
1) Powers
2) Martial Arts and any maneuvers
3) Cyber/bioware
4) Armor
5) Melee and Ranged Weapons
6) Gear
7) Devices
8) Programs

**Shadowrun 5* character sheets**
Issue #3861 - Quantity of Thrown Weapons be shown in the Ranged Weapons table

Modified the sheets as follows
1) Modified printing of notes for martial Arts and maneuvers to match Notes sheet.
2) Modified the formatting of Ranged Weapons table to show the quantity (from Gear) instead of the Ammo fixed amount of 1.
3) Only print "Current Form" for a shapeshifter when a shifted form is being printed
4) Removed printing of notes from the Gear columns - even small notes can mess with the layout of the columns. These notes have been added to the Notes sheet.
Note: This includes notes from devices, that are already displayed in the Devices section.
5) Modified printing of Armor list to separate Equipped and other into separate blocks.

**Shadowrun 5 set CSS.xslt**
This is the file used by the Shadowrun 5* character sheets (and now the Notes sheet) to define the css attributes used when creating the sheets - underlining, bold text, etc.

The language specific files for de-de and pt-br have formats that differ from the base version - specifically to keep column titles in lower case.

The base file was originally copied to all the additional languages. When changes to the base file were made, they were not carried forward to the additional versions.

Made the following changes to the css definitions:
1) Copied the format changes from the base file to the de-de and pt-br versions.
2) Modified files for the other languages to import the base version so that only the base version needs to be modified, with changes copied to de-de and pt-br versions, reducing the amount of work needed when making format changes.
3) Added a comment to the base file as a reminder that any changes should be also made to the individual language versions.
4) Added a comment to the base file describing how a language specific version should be created if and when it is needed.

**Vehicles character sheet**
The formatting of Ranged Weapons section now shows the quantity instead of Ammo and a fixed amount of 1.
(The Vehicles sheet uses the common ranged weapons formatting file).

**Fancy Blocks character sheet**
Only print "Current Form" for a shapeshifter when a shifted form is being printed

**other changes**
1) xs.RangedWeapons.xslt - this is where the changes to the display of Thrown Weapons have been made.
2) Created xt.MartialArts.xslt to format notes from Martial Arts and any maneuvers that is used by the Notes and Shadowrun 5* character sheets.


Chummer/sheets/Shadowrun 5 set.xslt
```diff
@@ -1005,14 +1036,24 @@
...
+        <xsl:if test="gears/gear[iscommlink = 'True']
+             or armors/armor/gears/gear[iscommlink = 'True']
+             or cyberwares/cyberware/gears/gear[iscommlink = 'True']
+             or cyberwares/cyberware/children/cyberware/gears/gear[iscommlink = 'True']
+             or weapons/weapon/accessories/accessory/gears/gear[iscommlink = 'True']
+             or gears/gear[isprogram = 'True']
+             or armors/armor/gears/gear[isprogram = 'True']
+             or cyberwares/cyberware/gears/gear[isprogram = 'True']
+             or cyberwares/cyberware/children/cyberwear/gears/gear[isprogram = 'True']
+             or weapons/weapon/accessories/accessory/gears/gear[isprogram = 'True']">
```
DelnarErsike, line 1048: I see what you're trying to do, but these are awkward and also don't pass all use cases. The problem is nesting: cyberware can have many layers of parents (e.g. modular lower limbs attached to modular full limbs att), as can weapons (via underbarrels) and gear (via plugins). These conditions only catch the first layer. If you want to be thorough, you'd want to walk through every single possible descendent nodeset that can lead to a commlink. This should be possible in XSL because it's possible in XPath, I just don't know how off the top of my head.

Chummer/sheets/Notes set.xslt
```diff
@@ -238,6 +215,191 @@
...
+              <xsl:call-template name="MartialArtsList"/>
+            </table>
+          </div>
+          <xsl:call-template name="RowSummary">
+            <xsl:with-param name="text" select="$lang.MartialArts"/>
+            <xsl:with-param name="blockname" select="'MartialArtsBlock'"/>
+          </xsl:call-template>
+        </xsl:if>
+
+        <xsl:variable name="cyberlist" select="cyberwares/cyberware[notes != ''] or cyberwares/cyberware/children/cyberware[notes != '']"/>
```
DelnarErsike, line 235: Keep in mind, cyberware can theoretically nest an infinite number of times. You want a selection that gets every single descendant, not just the first children. This is most obvious when you start dealing with modular cyberware.

Chummer/sheets/Notes set.xslt
```diff
@@ -238,6 +215,191 @@
...
+              </xsl:for-each>
+            </table>
+          </div>
+          <xsl:call-template name="RowSummary">
+            <xsl:with-param name="text" select="$lang.Armor"/>
+            <xsl:with-param name="blockname" select="'ArmorBlock'"/>
+          </xsl:call-template>
+        </xsl:if>
+
+        <xsl:variable name="rangedlist" select="weapons/weapon[type = 'Ranged' and (notes != '' or accessories/accessory/notes != '' or mods/weaponmod/notes != '')]"/>
```
DelnarErsike, line 287: Keep in mind, weapons can theoretically nest an infinite number of times via underbarrels. You want a selection that gets every single descendant, not just the first children.

Chummer/sheets/Notes set.xslt
```diff
@@ -238,6 +215,191 @@
...
+              </xsl:for-each>
+            </table>
+          </div>
+          <xsl:call-template name="RowSummary">
+            <xsl:with-param name="text" select="$lang.MeleeWeapons"/>
+            <xsl:with-param name="blockname" select="'MeleeWeaponsBlock'"/>
+          </xsl:call-template>
+        </xsl:if>
+
+        <xsl:variable name="gearlist" select="gears/gear[iscommlink != 'True' and (notes != '' or children/gear/notes != '')]"/>
```
DelnarErsike, line 339: Keep in mind, gear can theoretically nest an infinite number of times, and gear parents (e.g. cyberware) can also sometimes nest an infinite number of times. You want a selection that gets every single descendant, not just the first children.

Chummer/sheets/Notes set.xslt
```diff
@@ -238,6 +215,191 @@
...
+             or armors/armor/gears/gear[iscommlink = 'True' and (notes != '' or children/gear/notes != '')]
+             or cyberwares/cyberware/gears/gear[iscommlink = 'True' and (notes != '' or children/gear/notes != '')]
+             or cyberwares/cyberware/children/cyberware/gears/gear[iscommlink = 'True' and (notes != '' or children/gear/notes != '')]
+             or weapons/weapon/accessories/accessory/gears/gear[iscommlink = 'True' and (notes != '' or children/gear/notes != '')]"/>
+        <xsl:variable name="programlist" select="gears/gear[isprogram = 'True' and (notes != '' or children/gear/notes != '')]
+             or armors/armor/gears/gear[isprogram = 'True' and (notes != '' or children/gear/notes != '')]
+             or cyberwares/cyberware/gears/gear[isprogram = 'True' and (notes != '' or children/gear/notes != '')]
+             or cyberwares/cyberware/children/cyberwear/gears/gear[isprogram = 'True' and (notes != '' or children/gear/notes != '')]
+             or weapons/weapon/accessories/accessory/gears/gear[isprogram = 'True' and (notes != '' or children/gear/notes != '')]"/>
```
DelnarErsike, line 377: See my comment about nesting gear and nesting gear parents in `Shadowrun 5 set.xslt`.

DelnarErsike (changes requested):
Looks fine, only 3 main points of feedback:

1. Why is it necessary to check for a `BOD` attribute having an attribute category in order to know if someone is a shifter or not? Ideally, the check should work regardless of what attributes are defined internally, especially if in the future, attribute definitions would be unhardcoded. This just seems like pointless technical debt to me. Not a showstopper though.
2. If all you're doing in a template file is importing another template, I see no reason to have the first template file. Just import the template you want directly? Not a showstopper though.
3. A lot of the notes and commlink detection stuff won't deal well with nested items, e.g. modular cyberlimbs or individual sensors inside a sensor array that's in some sensor housing. Ideally, we'd want to do a thorough job the first time around instead of coming back and trying to tweak stuff as complaints come in of people not seeing some of their notes in their individual, specific cases. This is a showstopper and the only feedback I think is necessary to address before merging.

Lochabar:
1) Checking for BOD
I'm trying to avoid the printing of the blank "Current Form" box for every character that is not a shape shifter. To me a blank box suggests something is missing.

Looking at character files of shape shifter I couldn't see another way of determining if a "shifted" form is being displayed. If you can suggest something I will use it.

2) Importing default CSS definitions
I've wondered why a separate file CSSS file was created for every language since it happened.

Changes to the default file have made that were not duplicated in each of the language versions. So I was trying to minimize the amount of maintenance needed for future changes.

I was trying to avoid modifying all of the Shadowrun 5* xsl (other than de-de and pt-br) files to change the import to the base folder from the local folder.

I will go ahead eliminate the css xslt files from the rest of the language folders.
I think I will also rename the "Shadowrun 5 set CSS.xslt" file as "xs.Chummer5CSS.xslt" to show it is for general use not just the Shadowrun 5* sheets.

3) Chummer/sheets/xt.RangedWeapons.xslt the "colspan" command is combining the "Ammo" and "Loaded" columns for Thrown Weapons that otherwise separate for ranged weapons

Here is a screenshot showing both formats.
Weapons

4) Sorting weapons - missed listing it

5) Layers of parents / modular cyberware / weapons / gear / etc.

I'll fix the other items you've identified then drive myself crazy trying to deal with the problem of nested children.

DelnarErsike (changes requested):
Showstopper is still present (the notes sections still cannot handle nested items beyond immediate children), so status remains unchanged.

Outsourcing of stylesheet to a single, proper one is a good start, but it would probably save a lot of headaches in the future if instead of having every single file import the default stylesheet, template overriding were to be used (so that the default stylesheet gets loaded with the core templates, but can get overridden by templates defined in language-specific sections for certain languages).

Lochabar:
I've moved the import of the CSS style sets template into the core style set and added special cases for German and Portuguese (Brazil).
I did not use the common definitions template for the Fancy Blocks and Vehicles sheets because there is so little overlap of the definitions.

I still need to work on the problem of multiple levels of notes in descendant definitions.
The current print of the first layer of gear notes probably covers the needs of most uses.

I would like to implement all these changes and then work on the layers of notes separately.

DelnarErsike (approved):

PR #4316, opened 2021-02-15
Removed "Trigger Removal" applying twice, improved some translations, added missing German data changes

JoschiGrey:
Rulebook statblocks assume, that all accessories are installed. Therefore Chummer applied trigger removal twice. This was fixed by reducing the stated accuracy of all guns affected guns by one.

The optional rule "AlternateMetatypeAttributeKarma" description was completely unreadable in German and now accurately states it's function.

The cost of "Perceptive Defender" is 3 instead of 4 in the German version of run and gun. This was added to the "German Data Changes - Qualities".

An alternate rule interpretation for the in debt quality from the German rules was interpreted as German Data Change.

P.s. If I made some formal error while creating this pull request please let me know, I'm not very experienced with Git.


DelnarErsike (approved):
In Debt change is unnecessary because BP contribution is the default, otherwise good.

JoschiGrey:
ok, thanks for the merge.

Guess it was an old version in which the default was different, didn't double check if something changed on that part.
Will do next time I find some small stuff to contribute!

PR #4383, opened 2021-04-14
Sustained spell tracker

ArchonMegalon:


ArchonMegalon:
@DelnarErsike @chummer5a @joha4270 

Here it is! The merge to discuss a new feature from @JoschiGrey

ArchonMegalon:
Here the first questions:

I used a flow Panel and a new Control element for visualising the sustained spells.
The spells are stored in an ObservableCollection as Spells, at the moment with all the information that a saved spell normally has.

But if I now sustained the same spell 2 times, .Remove always removes the first instance of the spell and not the one I clicked at (because they are identical at the moment)

I Planned to add a Spell.Force to Spells that just isn't saved loaded or present in the normal character.file spells or the spell xml.
That would allow to Save the force of an sustained spell and delete the correct one (unless again 2 are identical but then it just doesn't really matter I guess).

I'm however unsure if just adding this to spells would somehow interfere with normal spells.

Also could you explain to me in broad strokes how "Improvements" work?
I never done anything T4 templates. I planned to handle psyche with it as suggested.
C#
//Deletes a sustained spell from the Sustained Spell Collection

        private void UnsustainSpell(object sender, EventArgs e)
        {
            if (sender is SustainedSpellControl objSender)
            {
                Spell objSustainedSpell = objSender.SustainedSpellObject;
                if (!CommonFunctions.ConfirmDelete(LanguageManager.GetString("Message_DeleteSustainedSpell")))
                    return;

                CharacterObject.SustainedSpells.Remove(objSustainedSpell);

                IsCharacterUpdateRequested = true;
                IsDirty = true;
            }
        }
This updates the displayed control elements.

`case NotifyCollectionChangedAction.Remove:
                        {
                            foreach (Spell objSustainedSpell in notifyCollectionChangedEventArgs.OldItems)
                            {
                                int intMoveUpAmount = 0;

                                if (panSustainedSpells == null)
                                    continue;

                                int intSustainedSpells = panSustainedSpells.Controls.Count;

                                for (int i = 0; i < intSustainedSpells; ++i)
                                {
                                    Control objLoopControl = panSustainedSpells.Controls[i];
                                    if (objLoopControl is SustainedSpellControl objSustainedSpellControl && objSustainedSpellControl.SustainedSpellObject == objSustainedSpell)
                                    {
                                        intMoveUpAmount = objSustainedSpellControl.Height;
                                        panSustainedSpells.Controls.RemoveAt(i);
                                        objSustainedSpellControl.SpellDetailChanged -= MakeDirtyWithCharacterUpdate;
                                        objSustainedSpellControl.UnsustainSpell -= UnsustainSpell;
                                        objSustainedSpellControl.Dispose();
                                        i -= 1;
                                        intSustainedSpells -= 1;
                                    }
                                    else if (intMoveUpAmount != 0)
                                    {
                                        objLoopControl.Top -= intMoveUpAmount;
                                    }
                                }
                            }
                            if (panSustainedSpells.Controls.Count != CharacterObject.SustainedSpells.Count)
                            {
                                RefreshSustainedSpells(panSustainedSpells);
                            }
                        }`

Sorry for the multitude of messages, the code just didn't fit in and I suspected you might wanna take a look.

JoschiGrey:
I now resolved the problems that I raised in question one by not saving the exact spell into the ObservableCollection, but creating new spell objects with spell.create.
I suspect this is less performant, then before.
Maybe just creating a new ID would do the trick that doesn't involve loading and reading the spells.xml. But it's read only isn't it?

chummer5a:
> But if I now sustained the same spell 2 times, .Remove always removes the first instance of the spell and not the one I clicked at (because they are identical at the moment)

I'd suggest making a separate SustainedSpell class instead of modifying the base Spell class. Pass the base spell in as an object reference, add force and net hits. Save the internal id of the Spell when you save, on Load get the spell by checking for a spell with the ID. If found set your reference to that spell, otherwise throw an error. 

> improvement methods T4 template
You shouldn't need to touch that directly, it's just for generating the enum list of improvement methods. If you're adding a new improvement type, add it in clsImprovement.cs public enum ImprovementType. The methods for new improvement blocks are then added in clsImprovement.cs\AddImprovementCollection.cs. The method you create is then responsible for generating an improvement object based on the input of the XML node that you provide it with. 

What I would suggest with the improvement management is that you take a look at the RefreshRedlinerImprovements method in clsCharacter.cs. That code checks for the number of relevant limbs for Redliner, then removes and replaces the relevant improvement with the appropriate value.

JoschiGrey:
> I'd suggest making a separate SustainedSpell class instead of modifying the base Spell class. Pass the base spell in as an object reference, add force and net hits. Save the internal id of the Spell when you save, on Load get the spell by checking for a spell with the ID. If found set your reference to that spell, otherwise throw an error.

That would make it all a bit cleaner I and probably easier to expand upon. It works right now, because spell.create assigns new internal IDs which makes every saved object unique. That prevents weird DataBindings and deletions.
Creating a new class with the original spell as object ref would probably be better in regards of performance too, since it doesn't need to create something from the xml. Gonna take a look at that and how references work.

Allowing to save the net hits is a neat idea, sadly available space is a big problem. The Spells and Spirits Tab is already quite crowded.

JoschiGrey:
Merged the new master into this branch, because it solved the loading and saving problems which hindered testing quite a bit.
Thanks for those fixes.

If someone could do the same for the chummer5a branch this is compared to to keep the file comparison clean, it would be much appreciated. @DelnarErsike

JoschiGrey:
So this is basically done, I suspect some code cleanup is required, but for that I would need your input.

To summarize everything.

**What has been added:**
- A new Class was added "SustainedSpells"
 -> This is a barebones version of the Spells class with some additional Properties.
 
- All known Spells can be sustained and are added to an Panel as Controls, similar to Spirits
 -> Net Hits for each spell can be tracked
 -> The Force of the spell can be tracked
 -> Spells are Self Sustained by default, but this can be controlled by a checkbox. (To allow tracking for spells in a Focus or 
      sustained by a Spirit)

- Skill Tooltips and Dice Pools are updated according to the amount of self sustained spells.
 -> The Drug Psyche can be controlled by an Checkbox which cuts sustaining Penalties in half.

**Stuff I'm not happy with right now:**
- Adding and Removing spells, as well as toggling self sustained status or psyche status feel quite slow, probably caused by the amount of recalculation for the tooltips and dice pools.

- My implementation of when to recalculate the Sustaining Penalty feels a bit clunky.
 -> For SelfSustained and Psyche it just has an CharacterObject.OnPropertyChanged("SustainingPenalty") in the Property
 -> The same OnPropertyChanged gets called after each ObservableCollection changed event.

- There is no support for custom Sustaining Penalty modifications.
 -> I'm not aware of any need for this from the top of my head, but if there is an improvement has to be added for that in the 
      future.

- The Position of the Psyche Checkbox is a bit weird.
 -> I originally intended it to be right above all other Controls in the flow panel. This could still be done, but requires some 
      refactoring of the RefreshSustainedSpells() Method. Some feedback in that regard (Do it or leave it where it's at right now), 
      would be much appreciated.

**For the Future:**
- There is no support for Technomancer Complex Forms right now.
 -> I honestly forgot about them in the beginning. It probably wouldn't be a big deal to implement this, but before I do that I 
     wanted to hear some feedback from you guys. 
- I translated some of the added Labels into German (I will finish that for sure in the near future), but that exhausts my 
  knowledge of languages. So I won't be able to introduce any other Translations, unless I start google translating.

DelnarErsike:
Gah, GitHub messed up stuff and now this got merged into a new branch instead of just adopting my upstream changes. I'll see if I can set something up...

DelnarErsike:
Continuing this in #4389 because GitHub won't let me do fancy stuff with committing permissions.

PR #4389, opened 2021-04-16
Sustained spell tracker

DelnarErsike:
Alright @JoschiGrey @ArchonMegalon @chummer5a @joha4270 , I guess this pull request is where we're tracking this branch now?

So, first and possibly stupid question: if you're going to make `SustainedSpell` a new class that's almost an exact copy of `Spell` but with some extra bits added on, why not make it inherit `Spell` and only specify the extra stuff that `SustainedSpell` adds? And then flag the stuff that's different with `virtual` in the original class?


JoschiGrey:
> So, first and possibly stupid question: if you're going to make SustainedSpell a new class that's almost an exact copy of Spell but with some extra bits added on, why not make it inherit Spell and only specify the extra stuff that SustainedSpell adds? And then flag the stuff that's different with virtual in the original class?

Well after briefly going over the dotnet article about inheritance, that would have been a great idea. If I had known about this.
This is the obvious stuff I miss, because I'm learning C#/object oriented programming while / with this project.
I'm sorry for extra work this is adding in review.

DelnarErsike:
Note: previous comments were in #4383.

DelnarErsike:
There are other options, of course.
You could set up an interface, e.g. `ISustainable`, alongside a similar child of `ComplexForm`, to extend sustaining stuff to Complex Forms. It might even make sense in that case to have the ObservableCollection of sustained spells be turned into an `ISustainable` ObservableCollection.

For something like this, it's definitely useful to have good knowledge about object-oriented design and how it works (at least in the case of C#, functional programming is a whole different beast). Not because it's strictly better or worse, mind you, but because it will save you a lot of headaches and tedious work in the future... assuming you do it well. If you do it poorly, it'll give you even more headaches. So if you want to do fancy stuff, make sure you both understand it first and do it for a reason that isn't just about being fancy.

JoschiGrey:
So to make sure I understood the advantage of an Interface correctly in this scenario.

You suggested an `ISustainable` Interface.
This could contain some common properties like `PsycheActive`, `Force` and `NetHits`.
This Interface can be be implemented into our different classes that could be sustained like `Spell` or `ComplexForm`.

We then could us a `ObservableCollection<ISustainable>`, that could hold and handle all objects that `ISustainable` was implemented in.
That saves us the work of writing new `Create()` Methods to convert each new object type we maybe want to add later for some reason into an `SustainedSpell`.
(For Example `AdaptPowers` or `CritterPowers`).
In that case we could also easily reuse the naming methods and so on without needing to copying them over to `SustainedSpell` (Because we don't have a SustainedSpell anymore and don't convert object types)?

DelnarErsike:
Depends on how you wish to implement it.

You could implement it how you mentioned, by putting common properties of sustained spells and complex forms into their respective classes, connecting them up with an `ISustainable` interface, and then manipulating sustained "stuff" in the program through interactions with `ISustainable` objects instead of having separate logic for spells and complex forms.

You could also implement different derived classes, `SustainedSpell`, `SustainedComplexForm`, `SustainedCritterPower`, etc., connect those up with `ISustainable`, and then interact with sustained stuff primarily though `ISustainable`s (or maybe `ISustainedCastable` might be a better name in this case, I don't know).

Interfaces and inheritance is purely about making things easier for the programmer, they don't really have a speed benefit compared to just writing everything out manually (and doing it well). So whatever design you think would be the easiest to deal with and the easiest to understand, go with that one, even if it looks a lot more primitive than the fanciest stuff you could cook up with C#.

JoschiGrey:
Thanks for the explanation! I'm gonna test around a bit.

As I see it right now both implementations raise different advantages and disadvantages.

Creating new derived classes for every object type, seems easier to read and keeps the base classes cleaner. It also feels easier to implement / set up for the first time.

Just interfacing the base classes creates less new classes / files and everything is a bit more concentrated. It also seems to be easier / faster to expand upon, but harder to convert to.

JoschiGrey:
This Pull Request can be closed by the way, the newest version is over at #4393 
I opened a new one that merges my own branch into master, to be able to push changes.

PR #4391, opened 2021-04-18
Update Translater

wakaba-github:
・Improvement points

translate tag (not attribute), this can now be changed.

・Remaining problems

translated cell and Changed column state is not updated immediately after the change
reselect section combo box , It is displayed correctly when it is redisplayed.

I test Japanese only , I'm sorry if there is a problem with other language.

### DIFF
--- a/Translator/frmTranslate.cs
+++ b/Translator/frmTranslate.cs
@@ -247,6 +247,20 @@ namespace Translator
                     if (xmlNodeLocal?.Attributes != null)
                     {
                         xmlNodeLocal.Attributes["translate"].InnerText = strTranslated;
+                        XmlAttribute objAttrib = xmlNodeLocal.Attributes?["translated"];
+                        if (objAttrib != null)
+                        {
+                            if (!flag)
+                                xmlNodeLocal.Attributes.Remove(objAttrib);
+                            else
+                                objAttrib.InnerText = bool.TrueString;
+                        }
+                        else if (flag)
+                        {
+                            objAttrib = _objDataDoc.CreateAttribute("translated");
+                            objAttrib.InnerText = bool.TrueString;
+                            xmlNodeLocal.Attributes.Append(objAttrib);
+                        }
                     }
                 }
                 else
@@ -272,6 +286,22 @@ namespace Translator
 
                             element.InnerText = strNameOnPage;
                         }
+                    }
+
+
+                    XmlAttribute objAttrib = xmlNodeLocal.Attributes?["translated"];
+                    if (objAttrib != null)
+                    {
+                        if (!flag)
+                            xmlNodeLocal.Attributes.Remove(objAttrib);
+                        else
+                            objAttrib.InnerText = bool.TrueString;
+                    }
+                    else if (flag)
+                    {
+                        objAttrib = _objDataDoc.CreateAttribute("translated");
+                        objAttrib.InnerText = bool.TrueString;
+                        xmlNodeLocal.Attributes?.Append(objAttrib);
                     }
                 }
             }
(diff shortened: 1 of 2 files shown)


global.json
```diff
@@ -1,6 +1,6 @@
 {
     "sdk": {
-        "version": "3.1.404",
+        "version": "5.0.201",
```
DelnarErsike, line 3: This should be undone; not everyone will have access to .NET SDK 5.0, and no part of Chummer really needs it.

Translator/frmTranslate.cs
```diff
@@ -239,28 +239,14 @@ private void dgvSection_CellValueChanged(object sender, DataGridViewCellEventArg
...
-                                xmlNodeLocal.Attributes.Remove(objAttrib);
-                            else
-                                objAttrib.InnerText = bool.TrueString;
-                        }
-                        else if (flag)
-                        {
-                            objAttrib = _objDataDoc.CreateAttribute("translated");
-                            objAttrib.InnerText = bool.TrueString;
-                            xmlNodeLocal.Attributes.Append(objAttrib);
-                        }
```
DelnarErsike, line 263: This needs to be rethought. The `translated` tag is there for if something's English name matches its translated name, which is the case sometimes in German and French. Since stuff without a translation gets highlighted in a different color, this makes sure those highlights don't happen when they aren't supposed to.

Translator/frmTranslate.cs
```diff
@@ -287,22 +273,6 @@ private void dgvSection_CellValueChanged(object sender, DataGridViewCellEventArg
...
-                            xmlNodeLocal.Attributes.Remove(objAttrib);
-                        else
-                            objAttrib.InnerText = bool.TrueString;
-                    }
-                    else if (flag)
-                    {
-                        objAttrib = _objDataDoc.CreateAttribute("translated");
-                        objAttrib.InnerText = bool.TrueString;
-                        xmlNodeLocal.Attributes?.Append(objAttrib);
-                    }
```
DelnarErsike, line 305: This needs to be rethought. The `translated` tag is there for if something's English name matches its translated name, which is the case sometimes in German and French. Since stuff without a translation gets highlighted in a different color, this makes sure those highlights don't happen when they aren't supposed to.

DelnarErsike (changes requested):
The `translated` tag is there for if something's English name matches its translated name, and it is used by some languages. I do not know why you want to remove it, because it is explicitly used by (and useful for) German and French.

global.json
wakaba-github: Sorry. I didn't know how to create the environment, so I left the settings for me.
This change will be undone.

Translator/frmTranslate.cs
wakaba-github: This part sets the items to be displayed in'translated'in the list display when'translate tag' is modified.
However,'translated' does not have in'ja-jp_data.xml', it is an item only for display.
If you reselect the section and redisplay it, it will be set correctly.
If you leave this part, the attribute "translated" will be registered in'ja-jp_data.xml'.
These should be unnecessary data in ja-jp_data.xml.
To fix it in its original form, it was necessary to save the changes in xml, redisplay the view, and select the column that was edited immediately before, but I could not incorporate it well.

Translator/frmTranslate.cs
wakaba-github: Another fix for the same reason.

Translator/frmTranslate.cs
DelnarErsike: Having `translated` attributes appear in `ja-jp_data.xml` should not be a problem though, right? And if the code is working the way it should, all it would be doing is creating and/or removing that attribute from items that are (or are not) translated, it should not affect the translations themselves.

wakaba-github: If'translated = true'is included in xml,'Translated' will not be changed as'True' unless you enter'False' in'Translated' after making the contents of'English' and'Text' the same. 
It sounds strange in behavior, but I think it's unlikely that you'll revert to'False', and if it doesn't affect the conversion in'Chummer', that's okay.

In my opinion, it's best not to include unnecessary data in xml
I hope someone will remove this unwanted attribute in the future.

DelnarErsike (approved):

PR #4393, opened 2021-04-19
Sustained spell tracker

JoschiGrey:
Yet another pull request for the sustained spell tracker, because it seems like it wasn't possible to give selective permissions for branches.

First PR for that was #4383 , second PR #4389.

Summary

**What has been added:**

The classes `SustainedSpell `and `SustainedComplexForm `got added
**-> They are both derived Classes from the respective base class**
**-> Those classes are interfaced with `ISustainable`**

All known Spells **and Complex Forms** can be sustained and are added to an Panel as Controls, similar to Spirits
-> Net Hits for each spell can be tracked
-> The Force of the spell can be tracked
-> Spells are Self Sustained by default, but this can be controlled by a checkbox. (To allow tracking for spells in a Focus or
sustained by a Spirit)

Skill Tooltips and Dice Pools are updated according to the amount of self sustained spells.
-> The Drug Psyche can be controlled by an Checkbox which cuts sustaining Penalties in half.
**-> The option to control the dice pool penalty per sustained spell was added as an house rule**

**Stuff I'm not happy with right now:**

After changing the Options for the penalty per sustained spell are changed the tooltips aren't updated immediately, because the event is always unsubscribed after loading/creating the Character in `clsCharacter`. I'm unsure why exactly, so I didn't mess with it. Could it be, that the second -= was meant as a += to resubscribe the event after the `CharacterOptions` got set successfully?

```c#
     public CharacterOptions Options
        {
            get => _objOptions;
            private set // Private to make sure this is always in sync with GameplayOption
            {
                if (_objOptions != value)
                {
                    if (_objOptions != null)
                        _objOptions.PropertyChanged -= OptionsOnPropertyChanged;
                    _objOptions = value;
                    if (_objOptions != null)
                        _objOptions.PropertyChanged -= OptionsOnPropertyChanged;
                    if (!IsLoading)
                        OnPropertyChanged();
                }
            }
        }
```


ArchonMegalon:
As I would see it, yes - that is probably an error and the second operator should be a +=.

Also: if you are unhappy with the slow ToolTip-(Re-)-Generation, why not make the OptionsOnPropertyChanged Async? When Handling the UI-Options you can always do it through the MainForm.DoThreadSave()-function.

(not that I have time right now to actually look at the code, so take this with a grain of salt).

chummer5a:
You shouldn't need specific permissions on the main chummer5a branch to update your pull request, just commit to https://github.com/JoschiGrey/chummer5a/tree/Sustained-Spell-Tracker and it should pick it up.

JoschiGrey:
No the problem is/was that the other Pull Request was from Delnar and wanted to merge https://github.com/chummer5a/chummer5a/tree/Sustained-Spell-Tracker into Master, not my own Branch.

DelnarErsike:
Note: the initial thing that led to the weird situation with a new branch being created was caused by me wanting to upstream merge `master` into the sustained spell pull request, as a lot of unmerged changes had piled up that polluted the actually file diffs.

JoschiGrey:
That's what I just realized after reading the invoke documentation again and checking it myself in debug.

I just stopped the main thread from locking up and the UI from being unresponsive. And thus it isn't faster, just feels better.
Gonna try your suggestion in a bit.

DelnarErsike:
If you want to speed up responsiveness, you can incorporate (more) caching and/or only perform the field updates when necessary. You can also do the two string builds in parallel if they do not change the internal state of the skill, which may be faster. If you cannot use Chummer while the items are calculating, don't forget to wrap the entire method's contents inside a CursorWait pattern so that the user knows why the program might not be responding.

ArchonMegalon:
I kind of lost track now regarding this commit. Would you say it is ready to be merged and we should take a final look at it?

JoschiGrey:
It should be ready to be merged, after you guys took a look at the code.
It does everything you need, besides tracking critter powers.
I was thinking about including it, but since that function is probably not as important, I was planning on doing it in the future.

What still needs to be done before the merge is adding documentation, regarding new strings and the changelog.

JoschiGrey:
TODO after this merge

- [ ] Rework loading in the async core
- [ ] add `RefreshingSustainingPenalty()` to the `PostLoadMethod` que

Chummer/frmCareer.cs
```diff
@@ -16313,6 +16368,295 @@ private void AddCyberwareSuite(Improvement.ImprovementSource objSource)
...
+        /// <param name="notifyCollectionChangedEventArgs"></param>
+        private void RefreshSustainedSpells(Panel flpSustainedSpells, Panel flpSustainedComplexForms, Panel flpSustainedCritterPowers, NotifyCollectionChangedEventArgs notifyCollectionChangedEventArgs = null)
+        {
+            if (flpSustainedSpells == null && flpSustainedComplexForms == null && flpSustainedCritterPowers == null)
+                return;
+
+            if (CharacterObject.SustainedCollection.Count != 0 && !chkPsycheActive.Visible)
+            {
+                //Sets up the Psyche Active Checkbox
+                chkPsycheActive.Visible = true;
```
DelnarErsike, line 16385: Setting this checkbox visible here and then not-visible later will cause flickering. You should find a different way of updating this checkbox's visibility.

Chummer/frmCareer.cs
```diff
@@ -16313,6 +16368,295 @@ private void AddCyberwareSuite(Improvement.ImprovementSource objSource)
...
+
+
+                    if (objSustained is SustainedSpell)
+                        flpSustainedSpells.Controls.Add(objSustainedSpellControl);
+
+                    if (objSustained is SustainedComplexForm)
+                        flpSustainedComplexForms.Controls.Add(objSustainedSpellControl);
+
+                    if (objSustained is SustainedCritterPower)
+                        flpSustainedCritterPowers.Controls.Add(objSustainedSpellControl);
```
DelnarErsike, line 16414: So this can potentially cause some weirdness if a class somehow inherits from multiple sustained object types. You should probably put `else`s before the latter `if`s.

Chummer/frmCareer.cs
```diff
@@ -16313,6 +16368,295 @@ private void AddCyberwareSuite(Improvement.ImprovementSource objSource)
...
+
+                                if (objSustained is SustainedCritterPower)
+                                {
+                                    objSustainedSpellControl.Top = intSustainedCritterPowers * objSustainedSpellControl.Height;
+                                    flpSustainedCritterPowers.Controls.Add(objSustainedSpellControl);
+                                }
+
+                                intSustainedComplexForms += 1;
+                                intSustainedSpells += 1;
+
```
DelnarErsike, line 16460: You forgot to update `intSustainedCritterPowers`.

Chummer/frmCareer.cs
```diff
@@ -16313,6 +16368,295 @@ private void AddCyberwareSuite(Improvement.ImprovementSource objSource)
...
+            if (treSpells.SelectedNode != null && treSpells.SelectedNode.Level > 0 && treSpells.SelectedNode.Tag is Spell objSpell)
+            {
+                SustainedSpell objNewSustainedSpell = new SustainedSpell(CharacterObject);
+
+                objNewSustainedSpell.Create(objSpell);
+
+                CharacterObject.SustainedCollection.Add(objNewSustainedSpell);
+
+                IsCharacterUpdateRequested = true;
+                IsDirty = true;
```
DelnarErsike, line 16620: These should be stuck into `SustainedSpellCollectionChanged` somehow, otherwise the program won't properly refresh itself if a sustained object is deleted outside of this method. That's why those events are set up in the first place.

Chummer/frmCareer.cs
```diff
@@ -16313,6 +16368,295 @@ private void AddCyberwareSuite(Improvement.ImprovementSource objSource)
...
+            if (treComplexForms.SelectedNode != null && treComplexForms.SelectedNode.Level > 0 && treComplexForms.SelectedNode.Tag is ComplexForm objComplexForm)
+            {
+                SustainedComplexForm objSustained = new SustainedComplexForm(CharacterObject);
+
+                objSustained.Create(objComplexForm);
+
+                CharacterObject.SustainedCollection.Add(objSustained);
+
+                IsCharacterUpdateRequested = true;
+                IsDirty = true;
```
DelnarErsike, line 16638: These should be stuck into `SustainedSpellCollectionChanged` somehow, otherwise the program won't properly refresh itself if a sustained object is deleted outside of this method. That's why those events are set up in the first place.

Chummer/frmCareer.cs
```diff
@@ -16313,6 +16368,295 @@ private void AddCyberwareSuite(Improvement.ImprovementSource objSource)
...
+            if (treCritterPowers.SelectedNode != null && treCritterPowers.SelectedNode.Level > 0 && treCritterPowers.SelectedNode.Tag is CritterPower objCritterPower)
+            {
+                SustainedCritterPower objSustained = new SustainedCritterPower(CharacterObject);
+
+                objSustained.Create(objCritterPower);
+
+                CharacterObject.SustainedCollection.Add(objSustained);
+
+                IsCharacterUpdateRequested = true;
+                IsDirty = true;
```
DelnarErsike, line 16656: These should be stuck into `SustainedSpellCollectionChanged` somehow, otherwise the program won't properly refresh itself if a sustained object is deleted outside of this method. That's why those events are set up in the first place.

Chummer/lang/en-us.xml
```diff
@@ -482,6 +482,10 @@
       <key>String_Delete</key>
       <text>Delete</text>
     </string>
+    <string>
+      <key>String_DeleteShort</key>
+      <text>X</text>
+    </string>
```
DelnarErsike, line 488: No. Find a way to put an image into your button or control or whatever, then use the same image that's used in all other places for removing stuff (minus sign in a red circle). This is going to look extremely hacky and amateurish when used the way I think you're using it.

Chummer/Controls/SustainedSpellControl.cs
```diff
@@ -0,0 +1,92 @@
...
+            lblSustainedSpell.Text = _objSustainedSpell.DisplayName(GlobalOptions.Language);
+            nudForce.DoDatabinding("Value", _objSustainedSpell, nameof(_objSustainedSpell.Force));
+            nudNetHits.DoDatabinding("Value", _objSustainedSpell, nameof(_objSustainedSpell.NetHits));
+
+            //Prevents the user from setting  SelfSustained of an SustainedCritterPower to true, because sustained critterpowers don't cause any dice loss
+            if (_objSustainedSpell is SustainedCritterPower)
+            {
+                chkSelfSustained.Visible = false;
+                lblSelfSustained.Visible = false;
+            }
```
DelnarErsike, line 45: Would this not work better in the constructor where `InitializeComponent()` is already? `_objSustainedSpell` is not supposed to change after being set in the constructor anyway, so it's not like there's a risk of breaking data bindings by hiding these from the very beginning.

DelnarErsike (changes requested):
There's still lots of boilerplate that should be addressed, and some very hacky parallelization that should probably be moved to a separate branch instead of being mixed up in the sustained spells tracker. You should also probably read up a bit more on how the `OnPropertyChanged()` and PropertyChanged-based event systems are meant to work because the way it's done with self-sustaining spells and CFs here is not at all how it should be done.

Also, there are a couple of lines in here that definitely cause build errors.

Chummer/Interfaces/ISustainable.cs
```diff
@@ -0,0 +1,78 @@
...
+        /// </summary>
+        /// <param name="objWriter"></param>
+        void Save(XmlTextWriter objWriter);
+
+        /// <summary>
+        /// Returns the Display name of an given Spell or Complex Form
+        /// </summary>
+        /// <param name="strLanguage"></param>
+        /// <returns></returns>
+        string DisplayName(string strLanguage);
```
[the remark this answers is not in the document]
JoschiGrey: Yes they are, this allowed me to keep code cleaner and smaller in other cases, this for example.

```   
           foreach(ISustainable objSustained in SustainedCollection)
           {
               objSustained.Print(objWriter, strLanguageToPrint);
           }
```
Without implementing the `Print()` method in the Interface I would need to distinguish wich `obj` it is beforehand.

The `IHasName` Interface only has the `Name` property. And I needed `DisplayName()` for the same reason I needed the other methods in the interface.

By the way, why don't we use a newer version of C#? The "Default Interface Implementations" sound like an nice feature.

Chummer/Backend/Derived/SustainedCritterPower.cs
```diff
@@ -0,0 +1,124 @@
...
+            }
+            objNode.TryGetStringFieldQuickly("name", ref strName);
+            if (!objNode.TryGetGuidFieldQuickly("sourceid", ref guiSourceID))
+            {
+                XmlNode node = GetNode(GlobalOptions.Language);
+                node?.TryGetGuidFieldQuickly("id", ref guiSourceID);
+            }
+            objNode.TryGetInt32FieldQuickly("force", ref _intForce);
+            objNode.TryGetInt32FieldQuickly("nethits", ref _intNetHits);
+        }
```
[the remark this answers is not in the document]
ArchonMegalon: Since I think Joschi wrote about being new to inheritance just a small comment:

You could use the keyword "virtual" in the baseclass for SaveDerived. That's exactly what it is for...

Chummer/Interfaces/ISustainable.cs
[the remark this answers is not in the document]
DelnarErsike: > Without implementing the `Print()` method in the Interface I would need to distinguish wich `obj` it is beforehand.

You misunderstand. An interface *should* only contain the methods that it needs to function. A thing that is sustainable does not necessarily have to be printable with a specific CultureInfo and string parameter, for example. What you're probably looking for is a common `IPrintable` interface. Right now, it might seem like needless work to implement `ISustainable` and `IPrintable` as two separate interfaces where all of your `ISustainable` stuff implements `IPrintable` anyway, but it makes sure that those two ways in which a class can function ("Can this be printed?" and "Does this have the properties of a thing that can be sustained?") are separated out and never inadvertently mixed in the future. A similar thing applies to `DisplayName()` stuff.

As for what to do about those right now, I'd probably flag those methods as `\\ TODO: Extract this out into a separate interface` for now, then when this pull request is merged, I'd immediately set up about rectifying that `TODO` in a new one.

> By the way, why don't we use a newer version of C#?

Because newer versions of C# require newer versions of .NET, and newer versions of .NET require newer versions of Windows. .NET Framework 4.7.2 is the most recent version still supported by Windows 7, and we have enough Windows 7 users that switching to a newer version of .NET should only be done if there really is something we can only do in the newer version. E.g., I had us switch from 4.5 to 4.7.2 when I needed 4.7.2's better DPI scaling support to handle a lot of DPI scaling issues.

> The "Default Interface Implementations" sound like an nice feature.

I see it as unnecessary when extension methods exist (and check out the `IHasMatrixAttributes` file to see said extension methods in action when working with an interface). To me, this "feature" really seems like just a more cleaned-up version of setting up an extension method with an interface-casted object as the `this` argument. Sure, the code is cleaner, but this is hardly a reason to switch to a .NET version that is not supported on Windows 7.

Chummer/frmCareer.cs
JoschiGrey: I originally didn't use `switch `s here, because I read they should only be used with >5 cases.

But instead using an `else` I now wrapped them into a switch with a `default`, because they are way easier to read and expand upon imo.
I also moved that one out into it's own `DetermineRefreshingPanle()` to reduce the size of this.

Chummer/UI/Skills/SkillControl2.cs
```diff
@@ -78,8 +79,8 @@ public SkillControl2(Skill objSkill)
 
                 btnAttribute.DoOneWayDataBinding("Text", objSkill, nameof(Skill.DisplayAttribute));
 
-                lblModifiedRating.Text = objSkill.DisplayOtherAttribute(_objAttributeActive.Abbrev);
-                lblModifiedRating.ToolTipText = objSkill.CompileDicepoolTooltip(_objAttributeActive.Abbrev);
+            var threadRefreshing = new Thread(SkillControl2_RefreshPoolTooltipAndDisplay) {IsBackground = true};
+            threadRefreshing.Start();
```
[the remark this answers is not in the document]
JoschiGrey: At the point I wrote this, I wasn't aware that UI Updates are the real problem.
I wanted to achieve two thing:

 - Performance
 - The UI not locking up every time one changed the `SelfSustained` or `PsycheActive` state.

In my personal bit of testing both worked out fine. 
Performance probably didn't got better in hindsight it just felt better. And `QueueThreadSafe` sounds like the solution to the second problem.

Will discard this entirely for now and add it to the TODO for later.

Chummer/Interfaces/ISustainable.cs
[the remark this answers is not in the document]
JoschiGrey: Not being able to support Win7 anymore is an good argument.

I already reduced the size of that interface by taking out `Print()`, because of a signature mismatch between `Spells` and the other two, after not overriding the complete `Spell.Print()` method anymore. `Load()` also kicked the bucket, because it was unnecessary in the first place.

Will mark the `DisplayName()` and `Save` as `TODO` for now.

JoschiGrey:
Lesson learned, don't amend commits you already pushed.
But it seems like everything worked out and all changes seem to be present in 9fcd691.

Most of the stuff you marked is done, mainly the Property Changers left.

JoschiGrey:
Everything you mentioned should be resolved now.

On the "After Merger ToDo":
- Remove all unnecessary stuff from `ISustainable `
- Take a look at the parallelization of `SkillControl2_RefreshPoolTooltipAndDisplay`

JoschiGrey:
So I just found two small omissions in `frmCareer`, that didn't set up `chkPsycheActiveTechnomancer`.

The puzzling part is, that as techno the tooltips don't update.
It works with Spells and everything is set up the same.

`RefreshSustainingPenalty()` is called and `_intSustainingPenalty` is set by the Method.
After that for some reason `Skill.OnCharacterChanged()` isn't called.

I guess I didn't set up an `OnPropertyChanged`, but I don't understand why it works for Spells and not for ComplexSpells.

Gonna take a look at it tomorrow.

Chummer/Classes/clsCharacter.cs
```diff
@@ -106,7 +106,7 @@ public sealed class Character : INotifyMultiplePropertyChanged, IHasMugshots, IH
         private string _strGameNotes = string.Empty;
         private string _strPrimaryArm = "Right";
 
-        private static readonly string[] s_LstLimbStrings = { "skull", "torso", "arm", "leg" };
+        private static readonly string[] s_LstLimbStrings = {"skull", "torso", "arm", "leg"};
```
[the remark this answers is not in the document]
JoschiGrey: I see what you mean. Don't know where the fuck they all come from.
Guess I pressed some wrong hotkey and ReSharper went apeshit in this file, I really should have seen this before the corresponding Commit/Push.

JoschiGrey:
All Conflicts are resolved, merge looks to went well.

Chummer/UI/Skills/SkillControl2.cs
```diff
@@ -568,5 +568,23 @@ private void SkillControl2_DpiChangedAfterParent(object sender, EventArgs e)
...
         }
+
+
+        /// <summary>
+        /// Refreshes the Tooltip and Displayed Dice Pool. Can be used in another Thread
+        /// </summary>
+        private void SkillControl2_RefreshPoolTooltipAndDisplay()
+        {
+            //TODO: Run this async somehow, QueueThreadSafe maybe.
+            if (_blnLoading || lblModifiedRating.Text != 0.ToString(GlobalOptions.CultureInfo))
```
DelnarErsike, line 579: This seems like an odd set of conditions. Are you sure that this code is how you want this entire method to work?

Chummer/Backend/Derived/SustainedComplexForm.cs
```diff
@@ -0,0 +1,184 @@
...
+        public event PropertyChangedEventHandler PropertyChanged;
+
+        #region Constructor, Create, Save, Load, and Print Methods
+        public SustainedComplexForm(Character objCharacter) : base(objCharacter)
+        {
+            //Create the GUID for new sustained Complex Forms
+            _guiID = Guid.NewGuid();
+            _objCharacter = objCharacter;
+
+            PropertyChanged += OnSustainedChanged;
```
DelnarErsike, line 62: Careful with this assignment. Because `OnSustinedChanged` refreshes character's sustaining penalties every time a property of this `SustainedComplexForm` is changed, things are going to go haywire whenever `CreateByMapping` is called,

Chummer/Backend/Derived/SustainedSpell.cs
```diff
@@ -0,0 +1,193 @@
...
+
+        #region Constructor, Create, Save, Load, and Print Methods
+
+        public SustainedSpell(Character objCharacter) : base(objCharacter)
+        {
+            //Create the GUID for new sustained spells
+            _guiID = Guid.NewGuid();
+            _objCharacter = objCharacter;
+
+            PropertyChanged += OnSustainedChanged;
```
DelnarErsike, line 62: Careful with this assignment. Because `OnSustinedChanged` refreshes character's sustaining penalties every time a property of this `SustainedSpell` is changed, things are going to go haywire whenever `CreateByMapping` is called,

Chummer/Backend/Uniques/Spell.cs
```diff
@@ -77,7 +78,7 @@ public Spell(Character objCharacter)
         /// <param name="blnExtended">Whether or not the Spell should be marked as Extended.</param>
         /// <param name="blnAlchemical">Whether or not the Spell is one for an alchemical preparation.</param>
         /// <param name="objSource">Enum representing the actual type of spell this object represents. Used for initiation benefits that would grant spells.</param>
-        public void Create(XmlNode objXmlSpellNode, string strForcedValue = "", bool blnLimited = false, bool blnExtended = false, bool blnAlchemical = false, Improvement.ImprovementSource objSource = Improvement.ImprovementSource.Spell)
+        public virtual void Create(XmlNode objXmlSpellNode, string strForcedValue = "", bool blnLimited = false, bool blnExtended = false, bool blnAlchemical = false, Improvement.ImprovementSource objSource = Improvement.ImprovementSource.Spell)
```
DelnarErsike, line 81: It doesn't seem like you need the `Create` method to be `virtual`?

Chummer/frmCareer.cs
```diff
@@ -16324,6 +16372,261 @@ private void AddCyberwareSuite(Improvement.ImprovementSource objSource)
...
+                    refreshingPanel = flpSustainedCritterPowers;
+                    break;
+
+                default:
+                    refreshingPanel = null;
+                    break;
+            }
+
+            return refreshingPanel;
+        }
```
DelnarErsike, line 16629: This method seems very specific and probably would be cleaner as a local method (i.e. a method that's defined inside another method) instead of even a private one.

DelnarErsike (changes requested):
Big thing that still needs changing is the way sustaining penalties are refreshed whenever a property of a `SustainedComplexForm` or `SustainedSpell` changes, and how that might also interact poorly with the way objects of those classes are created by mapping.

Chummer/Backend/Uniques/Spell.cs
JoschiGrey: Good catch, missed that one. Before I implemented `CreateByMapping()`, I had a `override Create()` in the derived classes.

Chummer/UI/Skills/SkillControl2.cs
JoschiGrey: The intention was, to only update the text, when it is necessary and for some reason I missed the obvious stupid bug I would have implemented...

Chummer/Backend/Derived/SustainedComplexForm.cs
JoschiGrey: So putting a breakpoint on `OnSustainedChanged()` shows it isn't called, when `CreateByMapping()` was called at the moment. Because the only property that fires(?) `OnPropertyChanged()` is `SelfSustained`, which isn't mapped over.

But to future prove it, I could add an `isCreating` bool and return `OnSustainedChanged` if that is true.

DelnarErsike: It would probably be easier to just move the `PropertyChanged` addition into `CreateByMapping` instead of having it in the constructor.

JoschiGrey: But not all instances of the class are created by mapping.
If you load a character they are just loaded from the `.xml`

DelnarErsike: Hmm... well, you could just *not* use events, or at least, not here. One option could be to assign refreshing of sustaining penalties on the `Character` side: every time an object is added to the appropriate `ObservableCollection`, it gets subscribed to an event where a change in `SelfSustained` causes the character to refresh their sustaining penalties. Another option could be to just trigger the given spell's character's `OnPropertyChanged` for their sustaining penalties whenever `SelfSustained` gets altered, but that's a bit hacky.

JoschiGrey: Triggering the `Character.OnPropertyChanged()` was what I've done before.

At the moment just throwing the `Character.RefreshSustainingPenalty()` into the set accessor of `SelfSustained` would have the exact same effect, just with less code, because then I don't need any `EventHandlers` in the first place.

So it seems like there is an `ObservableCollection<T>.PropertyChanged Event` ([dotnet](https://docs.microsoft.com/en-us/dotnet/api/system.collections.objectmodel.observablecollection-1.propertychanged?view=net-5.0)) and some solutions on how to listen to property changes of items in observable collections [example](https://www.codeproject.com/Tips/694370/How-to-Listen-to-Property-Chang). I haven't got the time to really look into that right now, but from the description it sounds like an possible solution.

Another way would be to just stick the assignment at the end of `Load()` and `CreateByMapping()`, that would also do the trick. However this just calls for mistakes and annoyances, as soon as someone touches the code later on.

DelnarErsike (approved):

PR #4408, opened 2021-05-01
Added cost and availability to underbarrel launcher and grapple gun

### LINKED ISSUE #4404: Underbarrel Grapple Launchers and Grenade Launchers are listed as 0 Availability and cost
**Describe the bug**
Underbarrel Grapple Launchers and Grenade Launchers are listed as 0 Availability and cost.

**To Reproduce**
1. Go to 'Weapons' subtab of 'Street Gear' tab
2. Click on 'Add Underbarrel Weapon' when right clicking any gun
3. Select either 'Underbarrel Grapple Gun' or 'Underbarrel Grenade Launcher'
4. Both will have 0 as their Availability and Cost

**Expected behavior**
Underbarrel Grapple Guns should have an Availability of 8R and a cost of 600; Underbarrel Grenade Launchers should have an Availability of 10F and a cost of 3500.

 - OS: Windows 10
 - Version 5.214.42

JoschiGrey:
Fixes #4404

### DIFF
--- a/Chummer/customdata/German Data Changes/custom_mentors.xml
+++ b/Chummer/customdata/German Data Changes/custom_mentors.xml
@@ -228,5 +228,42 @@
       <source>SG</source>
       <page>234</page>
     </mentor>
+
+    <mentor>
+      <id>7a837193-2b58-49be-9dbb-c1c0c108ca6e</id>
+      <name>Große Mutter (SG)</name>
+      <advantage>Allgemein: +2 Würfel für Proben auf erste Hilfe</advantage>
+      <disadvantage>Immer wenn ein Anhänger der Großen Mutter absichtlich oder unabsichtlich dem lebensspendenden Wesen der Natur (seiner Auslegung) zuwiderhandelt, erleidet er einen Würfelpoolmodifikator von -1 auf alle Handlungen, bis er seine Zuwiderhandlung wiedergutgemacht hat.</disadvantage>
+      <choices>
+        <choice>
+          <name>+2 to First Aid</name>
+          <bonus>
+            <specificskill>
+              <name>First Aid</name>
+              <bonus>2</bonus>
+            </specificskill>
+          </bonus>
+        </choice>
+        <choice set="2">
+          <name>+2 Würfel für Zauber, Rituale und Alchemische Erzeugnisse der Kategorie Heilung</name>
+          <bonus>
+            <spellcategory>
+              <name>Health</name>
+              <val>2</val>
+            </spellcategory>
+          </bonus>
+        </choice>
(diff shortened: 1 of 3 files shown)


JoschiGrey:
Pushed another fix/addition to this by accident instead of opening another PR.

Read the description of the second commit before merging and delete it from the PR if you disagree with my assessment.

DelnarErsike:
Uh... what is that fix in commit e699b25270f25229e64a91ac3736f7b7419dc446, @JoschiGrey?

You do realize that that results in two different "Great Mother" entries when using the German Data Changes, correct? One will be called "Great Mother" and the other will be called "Große Mutter (SG)"? This kind of duplication is exactly what that German Data Changes ruleset is meant to eliminate (as opposed to the previous solution of just adding fake sourcebooks that only contain German data).

JoschiGrey:
The German Street Grim implements a "Great Mother (Große Mutter)", which is not present in the English book.
It gives +2 on first aid, 2+ on health spells and "Empathic Healing".
The drawback translates more or less to: "If you hurt a lifegiving being you receive -1 on everything until you repent".

This book was released in 2015.


The English enhanced fiction novel "Sail Away, Sweet Sister" (SASS) introduced the Mentor Spirit "Great Mother"
It's description is nearly identical to the German SG Great Mother. With some key differences.
You choose between +2 on First aid or Medicine, you get +2 on health spells and you get "Rapid Healing".
The drawback gives -1 to all combat actions.

This book was released 2019 (and as it seems only in English)

Then there are the Great Mothers (FA) from the forbidden arcana which are identical in German and English.
They give +1 on Nature skills outside of urban areas, +2 on health spells and "Empathic Healing".
The drawback doubles background count from pollution or twisted elements.

Both got published 2017.

Now we got 3 Versions of "Great Mother" that all probably refer to the same mentor spirit to manage.

We could treat the FA one as an stealth errata for the German Great Mother (SG)
and the SASS version as an stealth errata to the English FA.

But what takes precedent as an German Data Change?
Does the German one supersede the later published English version?

Are those even stealth errata to the respective earlier published versions or did Catalyst just fuck up with SASS and Pegasus just translated the "new" FA mentor to German without realizing, they implemented this Mentor earlier.

At the moment the normal mentor.xml contains Great Mother (FA) and Great Mother (SASS).
And I added the GM(SG) that way we don't have to decide, which version is correct.

I agree this is a bit dumb, because we now got 3 versions of the Great Mother all with slightly different rules, from 3 different books if you use the German Data Changes.

---

Btw. I just realized the GM(SG) has the same ID as the GM(SASS) this has to be fixed. I'm sorry about that.

PR #4414, opened 2021-05-03
Fixed conflicting guid for mentro great mother (SG) and discussion how to handle this mentor.

JoschiGrey:
This Pull request contains the mentioned conflicting UID and is meant to discuss how this weird mentor spirit should be handled.

Following are the copy and pastes @DelnarErsike and my comments over in #4408

---
 **Joschi original description of e699b25**
The German Street Grim has the mentor "Große Mutter" (Great Mother). It is nearly identical to the one that is present in SASS, but not completely.

Since it is functionally a bit different and it's very unclear if they are meant as the same, adding it as another custom Mentor was the easiest way to include it, without having to make a rules call.

--- 
**Delnar**
Uh... what is that fix in commit e699b25, @JoschiGrey?

You do realize that that results in two different "Great Mother" entries when using the German Data Changes, correct? One will be called "Great Mother" and the other will be called "Große Mutter (SG)"? This kind of duplication is exactly what that German Data Changes ruleset is meant to eliminate (as opposed to the previous solution of just adding fake sourcebooks that only contain German data).

--------
**Joschi**
The German Street Grim implements a "Great Mother (Große Mutter)", which is not present in the English book.
It gives +2 on first aid, 2+ on health spells and "Empathic Healing".
The drawback translates More or Less to: "If you hurt a lifegiving being you receive -1 on everything until you repent".

This book was released in 2015.

The English enhanced fiction novel "Sail Away, Sweet Sister" (SASS) introduced the Mentor Spirit "Great Mother"
It's description is nearly identical to the German SG Great Mother. With some key differences.
You choose between +2 on First aid or Medicine, you get +2 on health spells and you get "Rapid Healing".
The drawback gives -1 to all combat actions.

This book was released 2019 (and as it seems only in English)

Then there are the Great Mothers (FA) from the forbidden arcana which are identical in German and English.
They give +1 on Nature skills outside of urban areas, +2 on health spells and "Empathic Healing".
The drawback doubles background count from pollution or twisted elements.

Both got published 2017.

Now we got 3 Versions of "Great Mother" that all probably refer to the same mentor spirit to manage.

We could treat the FA one as an stealth errata for the German Great Mother (SG)
and the SASS version as an stealth errata to the English FA.

But what takes precedent as an German Data Change?
Does the German one supersede the later published English version?

Are those even stealth errata to the respective earlier published versions or did Catalyst just fuck up with SASS and Pegasus just translated the "new" FA mentor to German without realizing, they implemented this Mentor earlier.

At the moment the normal mentor.xml contains Great Mother (FA) and Great Mother (SASS).
And I added the GM(SG) that way we don't have to decide, which version is correct.

I agree this is a bit dumb, because we now got 3 versions of the Great Mother all with slightly different rules, from 3 different books if you use the German Data Changes.

Btw. I just realized the GM(SG) has the same ID as the GM(SASS) this has to be fixed. I'm sorry about that. I can open a Pull Request for that as soon as we figured this out. This branch already got a new ID.

### DIFF
--- a/Chummer/customdata/German Data Changes/amend_mentors.xml
+++ b/Chummer/customdata/German Data Changes/amend_mentors.xml
@@ -17,11 +17,6 @@
     You can obtain the full source code for Chummer5a at
     https://github.com/chummer5a/chummer5a
 -->
-<!--
-  Hide the German skilljack because it's stupid
-  Set skilljack to flat 0.1 ess and R*1000 cost
-  Set skillwires to R*0.1 ess
--->
 <chummer>
   <mentors>
     <mentor>
@@ -46,6 +41,14 @@
           </bonus>
         </choice>
       </choices>
+    </mentor>
+    <mentor> <!--Great Mother SG-->
+      <id>a23c8a3d-b056-4173-88e4-e73bb64dd731</id>
+      <hide amendoperation="remove" />
+    </mentor>
+    <mentor> <!--Great Mother SASS-->
+      <id>7a837193-2b58-49be-9dbb-c1c0c108ca6e</id>
+      <hide />
     </mentor>
   </mentors>
 </chummer>
--- a/Chummer/customdata/German Data Changes/custom_mentors.xml
+++ b/Chummer/customdata/German Data Changes/custom_mentors.xml
@@ -228,41 +228,5 @@
       <source>SG</source>
       <page>234</page>
     </mentor>
-    <mentor>
-      <id>a23c8a3d-b056-4173-88e4-e73bb64dd731</id>
-      <name>Große Mutter (SG)</name>
-      <advantage>Allgemein: +2 Würfel für Proben auf erste Hilfe</advantage>
-      <disadvantage>Immer wenn ein Anhänger der Großen Mutter absichtlich oder unabsichtlich dem lebensspendenden Wesen der Natur (seiner Auslegung) zuwiderhandelt, erleidet er einen Würfelpoolmodifikator von -1 auf alle Handlungen, bis er seine Zuwiderhandlung wiedergutgemacht hat.</disadvantage>
-      <choices>
-        <choice>
-          <name>+2 to First Aid</name>
-          <bonus>
-            <specificskill>
-              <name>First Aid</name>
-              <bonus>2</bonus>
-            </specificskill>
-          </bonus>
-        </choice>
-        <choice set="2">
-          <name>+2 Würfel für Zauber, Rituale und Alchemische Erzeugnisse der Kategorie Heilung</name>
-          <bonus>
-            <spellcategory>
-              <name>Health</name>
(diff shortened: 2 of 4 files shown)


DelnarErsike:
Note that SASS was published before Street Grimoire or Forbidden Arcana, not in 2019. SASS in general also added a bunch of content that was later re-printed in other books, e.g. the One Less [Metatype] spell in Street Grimoire, the Inspired and College Education qualities in Run Faster, and Renfield in Hard Targets,

IMHO, the timeline was probably the following:

1. SASS printed and only in English. SASS has the first printing of Great Mother.
2. Street Grimoire is printed in English and German. In English, the One Less [Metatype] spell series is re-printed from SASS, but Great Mother isn't. In German, Great Mother is re-printed because SASS was never released in German, and Pegasus uses the opportuntiy to tweak it slightly.
3. Forbidden Arcana is printed with poor quality control and a bunch of stealth errata that is unclear if it is stealth errata or not (e.g. the removal of Enchanting from MysAds). Its re-printing of Great Mother is one such victim, and it is unclear if this version is meant as stealth errata or as an Alt along similar lines as Raven (Alt).

My advice would be to do the following:

* Have a German translation for Great Mother (SASS) as it appears in SASS, even though there is no German version for it.
* Have German Data Changes hide Great Mother (SASS) and add Great Mother (SG) as it appears in German Street Grimoire. It should have a new id and all text (except maybe the title) as English translations. Then, add a German "translation" (the actual, original text) to `de-de_data.xml`. The separate ids are needed to make sure the right German text is used depending on whether or not German Data Changes is activated.
* Optionally, we rename Great Mother (SASS) and Great Mother (SG) to just Great Mother, then rename FA's Great Mother to Great Mother (Alt). This is assuming that FA's Great Mother is to be interpreted as a separate mentor spirit and not stealth errata,... but Forbidden Arcana is Forbidden Arcana.

JoschiGrey:
Yeah it seems like the release date I saw on google was when it was published on google books or something.
But this timeline makes way more sense and handling SG as a errata to SASS seems fine to me.
I may not like it, because the German wording is atrocious, but it's a reasonable assumption. 

Renaming SASS/SG to Great Mother and FA to alt would be in line with the other FA alternative versions.

Just to be sure what you meant regarding the process.

Basically add SASS and SG to the mentor.xml as separate entities and translate both into German.
In `mentor.xml` add a `<hide />` to the SG version.

The German Data change then just removes the `hide` from the SG Version and adds it to the SASS version.

Because it isn't really clear and we got all the versions in English we could also make a data change
`Show all Great Mother Variants` that renames them to Great Mother (FA) / Great Mother (SASS) / Great Mother (SG) and shows them all three. At least that's what I'm gonna do for me and my players.

JoschiGrey:
This should be ready to merge now.

Issue #4416, opened 2021-05-05
Starting Nuyen from Lifestyle Should Let User Choose Which Lifestyle to Use


DelnarErsike:
Right now, Chummer's calculation of starting Nuyen from a character's lifestyle is based on the first lifestyle that the user added. This can be a problem if the character has multiple lifestyles, and their highest/preferred lifestyle is not the first one on their list.

Automatically finding and picking the best lifestyle will work with the RAW dataset, but Chummer has to be able to work well with custom rulesets that might add and/or change lifestyles such that there might be no objective best lifestyle for starting Nuyen. Because of this, the ideal solution should be to add the ability for the user to pick which lifestyle they want to use for rolling starting Nuyen and have the one with the highest average be the default option.

PR #4418, opened 2021-05-06
Selection of the lifestyle for starting  nuyen calc is now possible.

JoschiGrey:
Implements feature request #4416 

Added a combobox, that displays all purchased lifestyles.
The cbo is sorted by expected value, under the assumption, that the user rolled an average of 3.5/Dice.

### DIFF
--- a/Chummer/frmLifestyleNuyen.cs
+++ b/Chummer/frmLifestyleNuyen.cs
@@ -17,13 +17,17 @@
  *  https://github.com/chummer5a/chummer5a
  */
 ﻿using System;
+ using System.Collections.Generic;
+ using System.Linq;
  using System.Windows.Forms;
+ using Chummer.Backend.Equipment;
 
 namespace Chummer
 {
     public partial class frmLifestyleNuyen : Form
     {
         readonly Character _objCharacter;
+        private Lifestyle _objLifestyle;
 
         #region Control Events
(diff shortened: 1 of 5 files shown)


Chummer/frmLifestyleNuyen.cs
```diff
@@ -47,19 +51,47 @@ private void cmdOK_Click(object sender, EventArgs e)
...
+            RefreshCalculation(sender, e);
+            lblDice.Text = string.Format(GlobalOptions.CultureInfo, LanguageManager.GetString("Label_LifestyleNuyen_ResultOf"), SelectedLifestyle.Dice.ToString(GlobalOptions.CultureInfo));
+
+        }
+
+        private void RefreshCboSelectLifestyle()
+        {
+            List<Lifestyle> lstLifestyles = _objCharacter.Lifestyles.ToList();
+
+
```
SquadronROE, line 78: There a reason for this much white space?

SquadronROE (approved):
Tests pass locally and the changes look good.

Chummer/frmLifestyleNuyen.cs
JoschiGrey: I insert quite a lot of whitespace while writing, because that makes it easier for me somehow. Just forgot to clean up that part.

PR #4444, opened 2021-05-16
Support for language specific CustomData Manifests

### LINKED ISSUE #4263: Customdata folders should include a manifest
More info about what a customdata folder actually does is good. 
Add a manifest file that can be parsed as a tooltip or label when selecting the folder in options.

JoschiGrey:
Fixes #4263

This pull request adds support for language specific manifest files, that can explain the functionality of an Custom Data Folder in greater detail, than just the folder name.

- The contents of an manifest file are displayed as mouseover tooltip in the character options.
- Files need to be placed in an "Manifests" subfolder of the corresponding custom data directory and follow the naming convention `[Language Code.txt]` for example `en-us.txt`.
- An example can be found in the "German Data Changes" folder.
- The tooltip will default to `en-us` and display an additional massage if it has done so.
- The tooltip will display, that no manifest could be found. If it is missing, the file is empty, there isn't a "manifests" directory or it's inaccessible in some other way.

I thought about using different ways of handling languages, all got shortcomings in one form or another. A major one would have been using a .xml

Using a `manifest.xml` with languages as `<key>`
Pro:
- Easier to handle, because the files don't need to be in an specific directory and don't have to follow a naming convention

Con:
- Longer texts and formatting would require escape sequences, which is way more annoying

For me my solution seems easier to handle and people who implement custom data files already need to follow naming conventions.


DelnarErsike:
I kind of think an `.xml` file would still be better, mainly because:

1. Chummer would only need to read in one file to get all texts. You don't need to get the LanguageManager involved, either, just have each language's description be set based on an attribute in the encompassing node.
2. Manifests can contain more than just a description for the ruleset. For example, they could contain:
  * Names information (important if you might rename your directories)
  * Hashes and/or IDs for a unique, content-specific identifier
  * Version information
  * Dependency information (e.g. one ruleset depends on another)
  * Mutual exclusivity information
  * Authorship information

This data storage system is exactly the sort of thing that XML is meant for.

JoschiGrey:
If you consider those other use cases, I agree a `.xml` is way better.

I was thinking mostly about including longer texts, and writing those in an `.xml` is always incredibly annoying.
Some `string` magic, to include line breaks and such is probably also an option. This way you wouldn't need escape sequences and could just copy and paste a long string of text from some other editor into the `.xml`.

Reworking this to read in `.xml` won't be hard, gonna take a look at the possibility of automating the formatting a bit too.

At least functionality behind those other usages, will be a TODO for the future, but I like the idea of having dependencies (not sure if we really need them atm but the idea is nice).

---
"Negative Dependencies" or checks for different files, that all modify the same item in different ways could also be interesting as a feature for the future

JoschiGrey:
Ohh this push went bad... let me see if I can fix this somehow

PR #4445, opened 2021-05-16
Colored notes

Phil5555:
This PR adds an option to select a custom color for each note, so that one can use colors to their liking and to provide a quick overview of qualities, spells, gear, ...

Examples:
Mark all positive Qualities in green and negative ones in red
Mark all Fire-based spells in Orange
Have cars show their (current) color in the vehicle list
...

**Please review carefully before merging!** 
There where quite some code duplications in the original source and I'm not fully sure if I covered everything.


DelnarErsike (changes requested):
An interesting start, but there are definitely a few things that need to be rethought, even outside of technical implementations.

The main thing that would need to be rethought IMHO is the interaction of custom-colorable items with Light/Dark mode. Right now, all items with notes are set to the same color, but this color is automatically refreshed whenever the program switches between Light and Dark modes, making sure that text is always legible. If you've got custom colors that are cached directly, those will not be updated automatically if/when the program switches color modes or when a character is opened up in a mode different from when it was saved (can happen all the time with GMs opening up their players' characters). This can create all sorts of frustrations. E.g., a player in Dark Mode might set an item's color to something like `#e8f4f8` that is easily read on Dark Mode's dark backgrounds, but then when their GM opens up their save in Light Mode, that same color becomes incredibly hard to read on Light Mode's light backgrounds.

So, aside from the stuff I mentioned in my comments, there also needs to be a proper way to make sure that no matter what color a user selects, it gets automatically translated to a similarly legible one if/when the program is in a different mode than its current one. The way I'd approach it is to decide on a default mode (I'd go with Light Mode) and use the algorithm in `ColorManager` that is built to automatically switch colors between Light and Dark mode versions. When a user in Dark Mode saves a color for their character, that color is stored in its Light Mode equivalent, even if the color they see is the one they picked out for their own Dark Mode. Similarly, when a user in Dark Mode loads in a character, all of the save's custom colors get run through the same algorithm before being applied.

Phil5555:
First of all: Thanks for the review and the helpfull comments!

> when their GM opens up their save in Light Mode, that same color becomes incredibly hard to read on Light Mode's light backgrounds.

I considered it a bit confusing if a player tells their GM to look at the things colored in DarkRed and the GM is using a different theme and won't find it/look at other things.

This is why I stole the internal Dark/Light colors (via the private-->public change) and used them as background color. This was meant to serve as a hint how the color would look like an both modes.

But I see your point there. I'll rework it and see what it looks like than. Seems like it only changes the Saturation, so at least the Hue should be helpfull enough in other theme settings.

Would you be OK with opening up the
`private static Color GetDarkModeVersion(Color objColor)`
as a public function so it can be used from outside?

Or maybe even a
`public static Color GetCurrentThemeModeVersion(Color objColor)`
which internally checks if it should return the same Color (LightMode) or the changed one (DarkMode) and might even extend if there are more themes in the future?

Chummer/Character Creation/frmCreate.cs
```diff
@@ -4993,12 +4993,13 @@ private void tsVehicleGearNotes_Click(object sender, EventArgs e)
             {
                 case Gear objGear:
                 {
-                    using (frmNotes frmItemNotes = new frmNotes(objGear.Notes))
+                    using (frmNotes frmItemNotes = new frmNotes(objGear.Notes, objGear.NotesColor))
```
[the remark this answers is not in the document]
Phil5555: I didn't dare to change, but maybe it would be even more valuable to have a Notes class which can hold several things per object.
For now that would be
* the text
* the ForegroundColor

and might be extended to
* an icon
* the background color

DelnarErsike:
> Would you be OK with opening up the `private static Color GetDarkModeVersion(Color objColor)` as a public function so it can be used from outside?

If you had to open up a `ColorManager` function to `public`, I would much rather it be `GenerateDarkModeColor()`. The latter one calls `GetDarkModeVersion()` anyway, but then stores the color in a static dictionary so that it's only ever created once. This means that if you need to generate the Dark Mode versions of 20 different items that all have the same color, `GenerateDarkModeColor()` will only actually do the color creation and modification once. For the remaining 19 items, it will just fetch the already-generated color from its dictionary.

What is missing though is an inverse/reverse version of `GetDarkModeVersion()`. I.e. if someone sets a color in Dark Mode, you can pass that color into the function and it will do the exact inverse of `GetDarkModeVersion()`. You can double-check that it's an exact inverse by then passing the result back through `GetDarkModeVersion()` or `GenerateDarkModeColor()` and seeing if you get the same color as what you started with. Or, if you want, I could write that function and push it to the repo for you to (re)use.

Phil5555:
> What is missing though is an inverse/reverse version of GetDarkModeVersion()

Right. I tried if the existing one is by accident working. It's not:
```
_colNotes = colorDialog1.Color;
Color a = ColorManager.GenerateDarkModeColor(_colNotes);
Color b = ColorManager.GenerateDarkModeColor(a);
Color c = ColorManager.GenerateDarkModeColor(b);
Color d = ColorManager.GenerateDarkModeColor(c);
```
results in following colors:
Start: `#ff0000`
a( Dark): `#fbbcbc`
b (Light): `#e65c5c`
c (Dark): `#d87b7b`
d (Light): `#ba6c6c`

>  Or, if you want, I could write that function and push it to the repo for you to (re)use.
Would be great if you can provide something.

I still tend to have two functions, one which always returns the color in a standardized format that can be stored.
And one that takes a standardized version and returns a matching one for the current theme setting. So i don't have to care for this outside of the color manager.
But that part I can write myself, no worries.

Chummer/frmNotes.cs
```diff
@@ -34,10 +34,11 @@ public partial class frmNotes : Form
         private static int _intHeight = int.MinValue;
 	    private readonly bool _blnLoading;
         private string _strNotes;
+        private Color _colNotes;
 
         #region Control Events
-        public frmNotes(string strOldNotes)
-		{
+        public frmNotes(string strOldNotes, Color? colNotes = null)
```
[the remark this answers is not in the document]
Phil5555: I tried once more but actually `System.Drawing.Color` is a struct and has no conversion from null.
I tried to use Color.Empty as a default, but that's not a compile time constant and therefore also not allowed.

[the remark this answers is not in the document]
DelnarErsike: Hmm, then it's probably cleaner to have two constructors, one with a `Color` argument and one without.

Chummer/Backend/Equipment/ArmorMod.cs
```diff
@@ -1332,9 +1351,11 @@ public Color PreferredColor
             {
                 if (!string.IsNullOrEmpty(Notes))
                 {
+                    Color cGray = ColorManager.GrayText;
+                    Color NotesColorDimmed = Color.FromArgb((NotesColor.R + cGray.R) / 2, (NotesColor.G + cGray.G) / 2, (NotesColor.B + cGray.B) / 2);
```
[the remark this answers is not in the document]
Phil5555: As a improvement i have now added `public static Color GenerateCurrentModeDimmedColor(Color objColor)` which translates a given color to one with 20% less lightness (in Light mode) or 20% more lightness (in Dark mode)

There might be more tweaks needed, but it looks ok-ish.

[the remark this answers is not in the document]
DelnarErsike: Saturation. You'll mainly want to desaturate the color, not adjust its lightness towards 50%.

Compare the current set of colors for items with notes: `d2691e` normally, `d2b48c` if "grayed out". If you compare the two colors' HSV/HSB values (HSV/HSB is similar to HSL, but Value/Brightness is not tied as closely to Saturation the way Lightness is), they actually have the same Value/Brightness of 82 and very similar Hues. However, `d2691e` has a HSV Saturation of 86, whereas `d2b48c` has a HSV Saturation of 33.

Chummer/Backend/Uniques/Contact.cs
```diff
@@ -893,6 +900,15 @@ public string Notes
...
         }
 
+        /// <summary>
+        /// Forecolor to use for Notes in treeviews.
+        /// </summary>
+        public Color NotesColor
+        {
+            get => _colNotes;
+            set => _colNotes = value;
+        }
```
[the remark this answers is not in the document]
Phil5555: These are not working, but at least not breaking anything that i tested.
As they implement IHasNotes, I think I should at least keep the changes in even if it might be a bit confusing if there are no effects when choosing a color.

Phil5555:
if I use your inversion with a plain Red it gives me this series:

```
Color a = ColorManager.GenerateDarkModeColor(Color.Red);
Color b = ColorManager.GenerateInverseDarkModeColor(a);
Color c = ColorManager.GenerateDarkModeColor(b);
Color d = ColorManager.GenerateInverseDarkModeColor(c);
Color e = ColorManager.GenerateDarkModeColor(d);
Color r = ColorManager.GenerateInverseDarkModeColor(Color.Red);
```

a: `#fbbcbc`
b: `#f56d6d`
c: `#e97b7b`
d: `#dd7c7c`
e: `#bf6b6b`
r: `#ffffff`

Especially the last one would be quite confusing:
If a user is currently in Dark Mode and selects Red from the ColorPicker he would get a White text.

Am I using it wrong?

(I'll stop it for today, thanks so far!)

DelnarErsike:
The last one was used incorrectly (you tried to invert `#ff0000` directly instead of its Dark Mode conversion), but the others revealed a bug that I've now fixed in commit 16b37508b2750d48ce8fc0b1dcf5216a33ca7326. Basically, I was increasing Lightness, but then I didn't factor in the part where Saturation in HSL depends on Lightness, which would mess up inversions in an unpredictable way. I fixed the bug and tweaked the formulas to where things should work fine now.

Chummer/Backend/Equipment/ArmorMod.cs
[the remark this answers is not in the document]
Phil5555: I might need a little help there. Let me push what I have and maybe you can find a better approach to it

Chummer/frmNotes.cs
```diff
@@ -107,11 +107,13 @@ private void btnCancel_Click(object sender, EventArgs e)
...
         {
-            colorDialog1.Color = ColorManager.GenerateCurrentModeColor(_colNotes);
+            //colorDialog1.Color = ColorManager.GenerateCurrentModeColor(_colNotes);
+            _colNotes = colorDialog1.Color; //Selected color is always how it is shown in light mode, use the stored one for it.
             var resNewColor = colorDialog1.ShowDialog();
             if (resNewColor == DialogResult.OK)
             {
-                _colNotes = ColorManager.GenerateModeIndependentColor(colorDialog1.Color);
+                //_colNotes = ColorManager.GenerateModeIndependentColor(colorDialog1.Color);
+                _colNotes = colorDialog1.Color; //Selected color is always how it is shown in light mode.
```
Phil5555, line 116: This brings me back to the discussion for:
`Color r = ColorManager.GenerateInverseDarkModeColor(Color.Red);`

I think the following use case might be a bit confusing for a player:
The player has configured DarkMode. In the ColorSelection he selects White as the background color.
I store White as the version for Light Mode and actually show it as Black in Dark Mode.

--> I would think we require a function that is taking a Color that was selected in DarkMode and get back the related Color in LightMode.

Chummer/Classes/ColorManager.cs
```diff
@@ -316,6 +383,27 @@ private static Color InverseGetDarkModeVersion(Color objColor)
             return FromHsla(fltHue, fltNewSaturationHsl, fltNewLightness, objColor.A);
         }
 
+        private static Color GetBrightenedVersion(Color objColor)
+        {
+            // Built-in functions are in HSV/HSB, so we need to convert to HSL to invert lightness.
+            float fltHue = objColor.GetHue() / 360.0f;
+            float fltBrightness = objColor.GetBrightness();
+            float fltLightness = objColor.GetSaturation();
+            fltLightness = Math.Min(fltLightness + 0.2f, 1);
```
DelnarErsike, line 392: You sure you want to add 20 to the Saturation of the color instead of multiplying it by some value?

Chummer/Classes/ColorManager.cs
```diff
@@ -316,6 +383,27 @@ private static Color InverseGetDarkModeVersion(Color objColor)
...
+            return FromHsva(fltHue, fltBrightness, fltLightness, objColor.A);
+        }
+
+        private static Color GetDimmedVersion(Color objColor)
+        {
+            // Built-in functions are in HSV/HSB, so we need to convert to HSL to invert lightness.
+            float fltHue = objColor.GetHue() / 360.0f;
+            float fltBrightness = objColor.GetBrightness();
+            float fltLightness = objColor.GetSaturation();
+            fltLightness = Math.Max(0, fltLightness - 0.2f);
```
DelnarErsike, line 402: You sure you want to subtract 20 from the Saturation of the color instead of multiplying it by some value that is less than 1?

Chummer/Backend/Uniques/Spell.cs
```diff
@@ -102,6 +103,11 @@ public void Create(XmlNode objXmlSpellNode, string strForcedValue = "", bool bln
             }
             if (!objXmlSpellNode.TryGetMultiLineStringFieldQuickly("altnotes", ref _strNotes))
                 objXmlSpellNode.TryGetMultiLineStringFieldQuickly("notes", ref _strNotes);
+
+            String sNotesColor = ColorTranslator.ToHtml(ColorManager.HasNotesColor);
+            objXmlSpellNode.TryGetStringFieldQuickly("notesColor", ref sNotesColor);
+            _colNotes = ColorTranslator.FromHtml(sNotesColor);
+
```
DelnarErsike, line 110: Is this necessary in the `Create` method? Would there need to be data-definable colors for specific spells?

Chummer/Backend/Uniques/Quality.cs
```diff
@@ -199,6 +200,11 @@ public void Create(XmlNode objXmlQuality, QualitySource objQualitySource, IList<
             }
             if (!objXmlQuality.TryGetMultiLineStringFieldQuickly("altnotes", ref _strNotes))
                 objXmlQuality.TryGetMultiLineStringFieldQuickly("notes", ref _strNotes);
+
+            String sNotesColor = ColorTranslator.ToHtml(ColorManager.HasNotesColor);
+            objXmlQuality.TryGetStringFieldQuickly("notesColor", ref sNotesColor);
+            _colNotes = ColorTranslator.FromHtml(sNotesColor);
+
```
DelnarErsike, line 207: Is this necessary in the `Create` method? Would there need to be data-definable colors for specific qualities?

Chummer/Backend/Uniques/Metamagic.cs
```diff
@@ -75,6 +76,11 @@ public void Create(XmlNode objXmlMetamagicNode, Improvement.ImprovementSource ob
             objXmlMetamagicNode.TryGetInt32FieldQuickly("grade", ref _intGrade);
             if (!objXmlMetamagicNode.TryGetMultiLineStringFieldQuickly("altnotes", ref _strNotes))
                 objXmlMetamagicNode.TryGetMultiLineStringFieldQuickly("notes", ref _strNotes);
+
+            String sNotesColor = ColorTranslator.ToHtml(ColorManager.HasNotesColor);
+            objXmlMetamagicNode.TryGetStringFieldQuickly("notesColor", ref sNotesColor);
+            _colNotes = ColorTranslator.FromHtml(sNotesColor);
+
```
DelnarErsike, line 83: Is this necessary in the `Create` method? Would there need to be data-definable colors for specific metamagics?

Chummer/Backend/Uniques/MartialArtTechnique.cs
```diff
@@ -61,6 +62,11 @@ public void Create(XmlNode xmlTechniqueDataNode)
             if (xmlTechniqueDataNode.TryGetStringFieldQuickly("name", ref _strName))
                 if (!xmlTechniqueDataNode.TryGetMultiLineStringFieldQuickly("altnotes", ref _strNotes))
                     xmlTechniqueDataNode.TryGetMultiLineStringFieldQuickly("notes", ref _strNotes);
+
+            String sNotesColor = ColorTranslator.ToHtml(ColorManager.HasNotesColor);
+            xmlTechniqueDataNode.TryGetStringFieldQuickly("notesColor", ref sNotesColor);
+            _colNotes = ColorTranslator.FromHtml(sNotesColor);
+
```
DelnarErsike, line 69: Is this necessary in the `Create` method? Would there need to be data-definable colors for specific martial arts techniques?

Chummer/Backend/Uniques/MartialArt.cs
```diff
@@ -103,6 +104,11 @@ public void Create(XmlNode objXmlArtNode)
             objXmlArtNode.TryGetInt32FieldQuickly("cost", ref _intKarmaCost);
             if (!objXmlArtNode.TryGetMultiLineStringFieldQuickly("altnotes", ref _strNotes))
                 objXmlArtNode.TryGetMultiLineStringFieldQuickly("notes", ref _strNotes);
+
+            String sNotesColor = ColorTranslator.ToHtml(ColorManager.HasNotesColor);
+            objXmlArtNode.TryGetStringFieldQuickly("notesColor", ref sNotesColor);
+            _colNotes = ColorTranslator.FromHtml(sNotesColor);
+
```
DelnarErsike, line 111: Is this necessary in the `Create` method? Would there need to be data-definable colors for specific martial arts?

Chummer/Backend/Uniques/Enhancement.cs
```diff
@@ -70,6 +71,11 @@ public void Create(XmlNode objXmlArtNode, Improvement.ImprovementSource objSourc
             _objImprovementSource = objSource;
             if (!objXmlArtNode.TryGetMultiLineStringFieldQuickly("altnotes", ref _strNotes))
                 objXmlArtNode.TryGetMultiLineStringFieldQuickly("notes", ref _strNotes);
+
+            String sNotesColor = ColorTranslator.ToHtml(ColorManager.HasNotesColor);
+            objXmlArtNode.TryGetStringFieldQuickly("notesColor", ref sNotesColor);
+            _colNotes = ColorTranslator.FromHtml(sNotesColor);
+
```
DelnarErsike, line 78: Is this necessary in the `Create` method? Would there need to be data-definable colors for specific enhancements?

Chummer/Backend/Uniques/CritterPower.cs
```diff
@@ -80,6 +81,11 @@ public void Create(XmlNode objXmlPowerNode, int intRating = 0, string strForcedV
             _nodBonus = objXmlPowerNode.SelectSingleNode("bonus");
             if (!objXmlPowerNode.TryGetMultiLineStringFieldQuickly("altnotes", ref _strNotes))
                 objXmlPowerNode.TryGetMultiLineStringFieldQuickly("notes", ref _strNotes);
+
+            String sNotesColor = ColorTranslator.ToHtml(ColorManager.HasNotesColor);
+            objXmlPowerNode.TryGetStringFieldQuickly("notesColor", ref sNotesColor);
+            _colNotes = ColorTranslator.FromHtml(sNotesColor);
+
```
DelnarErsike, line 88: Is this necessary in the `Create` method? Would there need to be data-definable colors for specific critter powers?

Chummer/Backend/Uniques/ComplexForm.cs
```diff
@@ -79,6 +80,11 @@ public void Create(XmlNode objXmlComplexFormNode, string strExtra = "")
             objXmlComplexFormNode.TryGetStringFieldQuickly("fv", ref _strFV);
             if (!objXmlComplexFormNode.TryGetMultiLineStringFieldQuickly("altnotes", ref _strNotes))
                 objXmlComplexFormNode.TryGetMultiLineStringFieldQuickly("notes", ref _strNotes);
+
+            String sNotesColor = ColorTranslator.ToHtml(ColorManager.HasNotesColor);
+            objXmlComplexFormNode.TryGetStringFieldQuickly("notesColor", ref sNotesColor);
+            _colNotes = ColorTranslator.FromHtml(sNotesColor);
+
```
DelnarErsike, line 87: Is this necessary in the `Create` method? Would there need to be data-definable colors for specific complex forms?

Chummer/Backend/Uniques/Art.cs
```diff
@@ -70,6 +71,11 @@ public void Create(XmlNode objXmlArtNode, Improvement.ImprovementSource objSourc
             _objImprovementSource = objSource;
             if (!objXmlArtNode.TryGetMultiLineStringFieldQuickly("altnotes", ref _strNotes))
                 objXmlArtNode.TryGetMultiLineStringFieldQuickly("notes", ref _strNotes);
+
+            String sNotesColor = ColorTranslator.ToHtml(ColorManager.HasNotesColor);
+            objXmlArtNode.TryGetStringFieldQuickly("notesColor", ref sNotesColor);
+            _colNotes = ColorTranslator.FromHtml(sNotesColor);
+
```
DelnarErsike, line 78: Is this necessary in the `Create` method? Would there need to be data-definable colors for specific arts?

Chummer/Backend/Uniques/AIProgram.cs
```diff
@@ -73,6 +74,11 @@ public void Create(XmlNode objXmlProgramNode, string strExtra = "", bool boolCan
             objXmlProgramNode.TryGetStringFieldQuickly("page", ref _strPage);
             if (!objXmlProgramNode.TryGetMultiLineStringFieldQuickly("altnotes", ref _strNotes))
                 objXmlProgramNode.TryGetMultiLineStringFieldQuickly("notes", ref _strNotes);
+
+            String sNotesColor = ColorTranslator.ToHtml(ColorManager.HasNotesColor);
+            objXmlProgramNode.TryGetStringFieldQuickly("notesColor", ref sNotesColor);
+            _colNotes = ColorTranslator.FromHtml(sNotesColor);
+
```
DelnarErsike, line 81: Is this necessary in the `Create` method? Would there need to be data-definable colors for specific AI programs?

Chummer/Backend/Powers/Power.cs
```diff
@@ -161,6 +163,11 @@ public bool Create(XmlNode objNode, int intRating = 1, XmlNode objBonusNodeOverr
             _intRating = intRating;
             if (!objNode.TryGetMultiLineStringFieldQuickly("altnotes", ref _strNotes))
                 objNode.TryGetMultiLineStringFieldQuickly("notes", ref _strNotes);
+
+            String sNotesColor = ColorTranslator.ToHtml(ColorManager.HasNotesColor);
+            objNode.TryGetStringFieldQuickly("notesColor", ref sNotesColor);
+            _colNotes = ColorTranslator.FromHtml(sNotesColor);
+
```
DelnarErsike, line 170: Is this necessary in the `Create` method? Would there need to be data-definable colors for specific powers?

Chummer/Backend/Equipment/WeaponMount.cs
```diff
@@ -95,6 +96,11 @@ public void Create(XmlNode objXmlMod, decimal decMarkup = 0)
             objXmlMod.TryGetStringFieldQuickly("avail", ref _strAvail);
             if (!objXmlMod.TryGetMultiLineStringFieldQuickly("altnotes", ref _strNotes))
                 objXmlMod.TryGetMultiLineStringFieldQuickly("notes", ref _strNotes);
+
+            String sNotesColor = ColorTranslator.ToHtml(ColorManager.HasNotesColor);
+            objXmlMod.TryGetStringFieldQuickly("notesColor", ref sNotesColor);
+            _colNotes = ColorTranslator.FromHtml(sNotesColor);
+
```
DelnarErsike, line 103: Is this necessary in the `Create` method? Would there need to be data-definable colors for specific weapon mounts?

Chummer/Backend/Equipment/WeaponAccessory.cs
```diff
@@ -186,6 +187,10 @@ public void Create(XmlNode objXmlAccessory, Tuple<string, string> strMount, int
             if (!objXmlAccessory.TryGetMultiLineStringFieldQuickly("altnotes", ref _strNotes))
                 objXmlAccessory.TryGetMultiLineStringFieldQuickly("notes", ref _strNotes);
 
+            String sNotesColor = ColorTranslator.ToHtml(ColorManager.HasNotesColor);
+            objXmlAccessory.TryGetStringFieldQuickly("notesColor", ref sNotesColor);
+            _colNotes = ColorTranslator.FromHtml(sNotesColor);
+
```
DelnarErsike, line 193: Is this necessary in the `Create` method? Would there need to be data-definable colors for specific weapon accessories?

Chummer/Backend/Equipment/Weapon.cs
```diff
@@ -280,6 +285,11 @@ public void Create(XmlNode objXmlWeapon, ICollection<Weapon> lstWeapons, bool bl
             }
             if (!objXmlWeapon.TryGetMultiLineStringFieldQuickly("altnotes", ref _strNotes))
                 objXmlWeapon.TryGetMultiLineStringFieldQuickly("notes", ref _strNotes);
+
+            sNotesColor = ColorTranslator.ToHtml(ColorManager.HasNotesColor);
+            objXmlWeapon.TryGetStringFieldQuickly("notesColor", ref sNotesColor);
+            _colNotes = ColorTranslator.FromHtml(sNotesColor);
+
```
DelnarErsike, line 292: I think you've put this code in twice.

Chummer/Backend/Equipment/Weapon.cs
```diff
@@ -214,6 +215,10 @@ public void Create(XmlNode objXmlWeapon, ICollection<Weapon> lstWeapons, bool bl
             if (!objXmlWeapon.TryGetMultiLineStringFieldQuickly("altnotes", ref _strNotes))
                 objXmlWeapon.TryGetMultiLineStringFieldQuickly("notes", ref _strNotes);
 
+            String sNotesColor = ColorTranslator.ToHtml(ColorManager.HasNotesColor);
+            objXmlWeapon.TryGetStringFieldQuickly("notesColor", ref sNotesColor);
+            _colNotes = ColorTranslator.FromHtml(sNotesColor);
+
```
DelnarErsike, line 221: Is this necessary in the `Create` method? Would there need to be data-definable colors for specific weapons?

Chummer/Backend/Equipment/VehicleMod.cs
```diff
@@ -156,6 +157,10 @@ public void Create(XmlNode objXmlMod, int intRating, Vehicle objParent, decimal
             if (!objXmlMod.TryGetMultiLineStringFieldQuickly("altnotes", ref _strNotes))
                 objXmlMod.TryGetMultiLineStringFieldQuickly("notes", ref _strNotes);
 
+            String sNotesColor = ColorTranslator.ToHtml(ColorManager.HasNotesColor);
+            objXmlMod.TryGetStringFieldQuickly("notesColor", ref sNotesColor);
+            _colNotes = ColorTranslator.FromHtml(sNotesColor);
+
```
DelnarErsike, line 163: Is this necessary in the `Create` method? Would there need to be data-definable colors for specific vehicle mods?

Chummer/Backend/Equipment/Vehicle.cs
```diff
@@ -211,6 +212,10 @@ public void Create(XmlNode objXmlVehicle, bool blnCreateChildren = true)
             if (!objXmlVehicle.TryGetMultiLineStringFieldQuickly("altnotes", ref _strNotes))
                 objXmlVehicle.TryGetMultiLineStringFieldQuickly("notes", ref _strNotes);
 
+            String sNotesColor = ColorTranslator.ToHtml(ColorManager.HasNotesColor);
+            objXmlVehicle.TryGetStringFieldQuickly("notesColor", ref sNotesColor);
+            _colNotes = ColorTranslator.FromHtml(sNotesColor);
+
```
DelnarErsike, line 218: Is this necessary in the `Create` method? Would there need to be data-definable colors for specific vehicles?

Chummer/Backend/Equipment/LifestyleQuality.cs
```diff
@@ -155,6 +156,11 @@ public LifestyleQuality(Character objCharacter)
             objXmlLifestyleQuality.TryGetBoolFieldQuickly("contributetolimit", ref _blnContributeToLP);
             if (!objXmlLifestyleQuality.TryGetMultiLineStringFieldQuickly("altnotes", ref _strNotes))
                 objXmlLifestyleQuality.TryGetMultiLineStringFieldQuickly("notes", ref _strNotes);
+
+            String sNotesColor = ColorTranslator.ToHtml(ColorManager.HasNotesColor);
+            objXmlLifestyleQuality.TryGetStringFieldQuickly("notesColor", ref sNotesColor);
+            _colNotes = ColorTranslator.FromHtml(sNotesColor);
+
```
DelnarErsike, line 163: Is this necessary in the constructor? Would there need to be data-definable colors for specific lifestyle qualities?

Chummer/Backend/Equipment/Lifestyle.cs
```diff
@@ -170,6 +171,11 @@ public void Create(XmlNode objXmlLifestyle)
             objXmlLifestyle.TryGetBoolFieldQuickly("allowbonuslp", ref _blnAllowBonusLP);
             if (!objXmlLifestyle.TryGetMultiLineStringFieldQuickly("altnotes", ref _strNotes))
                 objXmlLifestyle.TryGetMultiLineStringFieldQuickly("notes", ref _strNotes);
+
+            String sNotesColor = ColorTranslator.ToHtml(ColorManager.HasNotesColor);
+            objXmlLifestyle.TryGetStringFieldQuickly("notesColor", ref sNotesColor);
+            _colNotes = ColorTranslator.FromHtml(sNotesColor);
+
```
DelnarErsike, line 178: Is this necessary in the `Create` method? Would there need to be data-definable colors for specific lifestyles?

Chummer/Backend/Equipment/Gear.cs
```diff
@@ -182,6 +183,10 @@ private void ChildrenOnCollectionChanged(object sender, NotifyCollectionChangedE
             if (!objXmlGear.TryGetMultiLineStringFieldQuickly("altnotes", ref _strNotes))
                 objXmlGear.TryGetMultiLineStringFieldQuickly("notes", ref _strNotes);
 
+            String sNotesColor = ColorTranslator.ToHtml(ColorManager.HasNotesColor);
+            objXmlGear.TryGetStringFieldQuickly("notesColor", ref sNotesColor);
+            _colNotes = ColorTranslator.FromHtml(sNotesColor);
+
```
DelnarErsike, line 189: Is this necessary in a method that gets called on/near creation of the gear? Would there need to be data-definable colors for specific gear?

Chummer/Backend/Equipment/Cyberware.cs
```diff
@@ -509,6 +510,10 @@ private void GearChildrenOnCollectionChanged(object sender, NotifyCollectionChan
             if (!objXmlCyberware.TryGetMultiLineStringFieldQuickly("altnotes", ref _strNotes))
                 objXmlCyberware.TryGetMultiLineStringFieldQuickly("notes", ref _strNotes);
 
+            String sNotesColor = ColorTranslator.ToHtml(ColorManager.HasNotesColor);
+            objXmlCyberware.TryGetStringFieldQuickly("notesColor", ref sNotesColor);
+            _colNotes = ColorTranslator.FromHtml(sNotesColor);
+
```
DelnarErsike, line 516: Is this necessary in a method on/near creation of gear? Would there need to be data-definable colors for specific gear?

Chummer/Backend/Equipment/ArmorMod.cs
```diff
@@ -140,6 +141,10 @@ public void Create(XmlNode objXmlArmorNode, int intRating, IList<Weapon> lstWeap
             if (!objXmlArmorNode.TryGetMultiLineStringFieldQuickly("altnotes", ref _strNotes))
                 objXmlArmorNode.TryGetMultiLineStringFieldQuickly("notes", ref _strNotes);
 
+            String sNotesColor = ColorTranslator.ToHtml(ColorManager.HasNotesColor);
+            objXmlArmorNode.TryGetStringFieldQuickly("notesColor", ref sNotesColor);
+            _colNotes = ColorTranslator.FromHtml(sNotesColor);
+
```
DelnarErsike, line 147: Is this necessary in the `Create` method? Would there need to be data-definable colors for specific armor mods?

Chummer/Backend/Equipment/Armor.cs
```diff
@@ -226,6 +227,10 @@ public void Create(XmlNode objXmlArmorNode, int intRating, IList<Weapon> lstWeap
             if (!objXmlArmorNode.TryGetMultiLineStringFieldQuickly("altnotes", ref _strNotes))
                 objXmlArmorNode.TryGetMultiLineStringFieldQuickly("notes", ref _strNotes);
 
+            String sNotesColor = ColorTranslator.ToHtml(ColorManager.HasNotesColor);
+            objXmlArmorNode.TryGetStringFieldQuickly("notesColor", ref sNotesColor);
+            _colNotes = ColorTranslator.FromHtml(sNotesColor);
+
```
DelnarErsike, line 233: Is this necessary in the `Create` method? Would there need to be data-definable colors for specific armor?

Chummer/frmNotes.Designer.cs
```diff
@@ -28,31 +28,125 @@ protected override void Dispose(bool disposing)
...
+            this.bufferedTableLayoutPanel2.ColumnCount = 2;
+            this.bufferedTableLayoutPanel2.ColumnStyles.Add(new System.Windows.Forms.ColumnStyle());
+            this.bufferedTableLayoutPanel2.ColumnStyles.Add(new System.Windows.Forms.ColumnStyle());
+            this.bufferedTableLayoutPanel2.Controls.Add(this.btnOK, 1, 0);
+            this.bufferedTableLayoutPanel2.Controls.Add(this.btnCancel, 0, 0);
+            this.bufferedTableLayoutPanel2.Location = new System.Drawing.Point(477, 290);
+            this.bufferedTableLayoutPanel2.Name = "bufferedTableLayoutPanel2";
+            this.bufferedTableLayoutPanel2.RowCount = 1;
+            this.bufferedTableLayoutPanel2.RowStyles.Add(new System.Windows.Forms.RowStyle());
+            this.bufferedTableLayoutPanel2.RowStyles.Add(new System.Windows.Forms.RowStyle(System.Windows.Forms.SizeType.Absolute, 29F));
```
DelnarErsike, line 105: You probably don't need two rows for this panel.

Chummer/frmNotes.cs
```diff
@@ -92,6 +94,34 @@ private void frmNotes_Shown(object sender, EventArgs e)
...
+            if (resNewColor == DialogResult.OK)
+            {
+                //_colNotes = ColorManager.GenerateModeIndependentColor(colorDialog1.Color);
+                _colNotes = colorDialog1.Color; //Selected color is always how it is shown in light mode.
+                updateColorRepresentation();
+            }
+        }
+        private void txtNotes_TextChanged(object sender, EventArgs e)
+        {
+            btnColorSelect.Enabled = txtNotes.Text.Length > 0;
```
DelnarErsike, line 122: I think there's a separate property that lets you get a text length directly? I'm not sure though. You can also check to see if the text is empty.

Chummer/frmNotes.cs
DelnarErsike: The idea you went with here is what I originally imagined would work:

* If the user is in Light Mode, the color they picked is stored directly as the color to be picked. When the color is shown, the original color is shown (also via `updateColorRepresentation`) in Light Mode and a generated Dark Mode version is shown in Dark Mode.
* If the user is in Dark Mode, the color they picked is assumed to be geared for Dark Mode. It first gets fed through `ColorManager.GenerateInverseDarkModeColor` to get the color in Light Mode that would turn into the selected one when in Dark Mode and store that. When the color is shown (also via `updateColorRepresentation`), the generated Light Mode color gets shown in Light Mode, while in Dark Mode, the generated Light Mode color has a Dark Mode version generated from it that gets shown. This generated Dark Mode version should end up being either exactly the same color as what the user originally picked in Dark Mode or very close to it.

Does this setup not work properly?

DelnarErsike (commented):
Little bits of extraneous code here and there, but it looks like the core functionality should work without major issues AFAICT. I'm honestly still a bit uncomfortable about the dimmed/brightened colors adding to Saturation instead of multiplying it, as that can end up doing some weird stuff, like making white look more reddish because of the way HSV/HSB color definitions work.

Chummer/frmNotes.cs
Phil5555: Ok, seems like we are aligned how it _should_ work. Therefore I have again commented in line 115 and removed 116.
For LightMode everything is easy, and I just need to store whatever was picked.

Now let us focus on DarkMode, because there it does not work as expected:
* User opens a color picker and selects Red (excuse my German Windows configuration)
grafik

* colorDialog1.Color is now set to ARGB(255, 255, 0, 0); `#ff0000`
* ColorManager.GenerateModeIndependentColor calls GenerateInverseDarkModeColor (because we are in DarkMode)
* GenerateInverseDarkModeColor() returns ARGB(255, 255, 255, 255) `#ffffff`
* I store this LightMode version internally, which is already the wrong color
* updateColorRepresentation now calculates DarkMode version from `#ffffff` which turns out to be ARGB (255, 64, 64, 64) `#404040`
* The text is not visible in dark mode:
grafik
grafik


--> GenerateInverseDarkModeColor seems to require changes. It also needs to accept colors like `#ff0000`, `#00ff00`, `#0000ff`

Chummer/Classes/ColorManager.cs
Phil5555: I'll change this to use a factor (1.15f). Result looks quite similar, but maybe I don't fully catch the corner cases you have in mind.

Chummer/Classes/ColorManager.cs
Phil5555: I'll change this to use a factor (0.85f). Result looks quite similar, but maybe I don't fully catch the corner cases you have in mind.

Chummer/frmNotes.Designer.cs
Phil5555: Indeed. The second row was not even visible in the graphical designed, seems like it's only the style. Will be removed in next revision.

Chummer/frmNotes.cs
Phil5555: Yes, that's `txtNotes.TextLength`. Will switch to that one with the next revision.

Chummer/Backend/Uniques/Spell.cs
Phil5555: Let's use this one to discuss also the other occurences.

Initially I was only trying to achive the same behaviour as with notes themselfs. Notes can occur in a data file and therefore I added the color section, too.

After a bit of consideration, I could imagine a real use case for this in the data-files. Especially when adding notes in data files, chosing a color that matches it's meaning (warning=yellow, default note = blue, ...) 

Personally I don't really care if we have it in here or not. If you like I can also remove it and it can be added later if a (better) use-case comes up.

DelnarErsike:
Alright, I tweaked the ColorManager's color conversion stuff in `master`, turns out .NET's `GetBrightness()` is actually Lightness, not Brightness, which massively tripped me up. Once the conflicts are sorted, we can merge this PR.

PR #4446, opened 2021-05-18
 Support for language specific CustomData Manifests, reworked

JoschiGrey:
I messed up somehow and included a bunch of Changes from Delnar, that really don't belong here and messed up the "files changed" of the pull request.
Original Pull request can be found here #4444 

This is a reworked version, which now takes and reads a `Manifest.xml`.
The schema definition `manifest.xsd` already contains a bunch of nodes for later use, which are unused at the moment.
They are also present as example in the `German Data Changes/manifest.xml` and marked as examples to prevent irritation in the future.

Only `<description>` is referenced in the code at the moment.

Languages are now handled by node attributes `<description lang="[lang code]">` and it will default to an node without any attributes.

The attribute less node is seen as the default and won't add any comments regarding missing languages if Chummer is set to English.


Chummer/customdata/German Data Changes/manifest.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<manifest>
+  <name>German Data Changes</name>
+  <guid>1ea1a21e-7cca-4ce9-b062-8f306880e8a6</guid>
+  <!--version and author are just here to serve as an example-->
+  <version>1.0</version>
+  <author>Multiple</author>
```
DelnarErsike, line 7: Hmm, maybe an `<authors>` node that can list any number of `<author>` children instead? The first one could be assumed to be the primary, or you could also allow a different one to be manually set as the primary author via attributes.

Chummer/customdata/Bone Lacing Adds to Body/manifest.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<manifest>
+  <name>Bone Lacing adds to Body</name>
+  <guid>66d452bf-f603-40ff-b792-08790975fcb7</guid>
```
DelnarErsike, line 4: Hmm, I don't know. GUIDs work when things are internal within Chummer, but I kind of feel like custom rulesets might need something more. I brought up hashes once as a kind of way to verify that two people are using the exact same contents of a custom ruleset, but I honestly don't know how I would integrate that into the manifest itself. Maybe have it just be a part of Chummer's checks for finding matching rulesets when loading `CharacterOptions`... I don't know. Could be a stupid idea to throw into the trash.

Chummer/customdata/manifest.xsd
```diff
@@ -0,0 +1,39 @@
...
+      <xs:element name="version" type="xs:string" minOccurs="0" maxOccurs="1"/>
+      <xs:element name="author" type="xs:string" minOccurs="0" maxOccurs="1"/>
+      <xs:element name="descriptions" minOccurs="0" maxOccurs="1">
+        <xs:complexType>
+          <xs:sequence>
+            <xs:element name="description" minOccurs="0" maxOccurs="unbounded"/>
+          </xs:sequence>
+        </xs:complexType>
+      </xs:element>
+      <!--Those don't have functionality yet, the idea for later is to use GuID to identify dependencies and exclusivities, maybe with the anyof/allof schema-->
```
DelnarErsike, line 22: Dependencies would probably need to work off of both an ID and a version number, not just an ID.

They could also work off of a hash from the contents of the other ruleset(s), but I already commented about how I have no idea how I would integrate that into the manifest. Even if it's calculated only within Chummer without the user needing to put it inside their manifest file, there's still the problem of how do you get the hashes of other rulesets into your dependency info for your own manifest without opening up Chummer and checking manually...

Chummer/Classes/clsXmlManager.cs
```diff
@@ -1857,6 +1857,69 @@ public bool Equals(CustomDataDirectoryInfo other)
...
 
+        /// <summary>
+        /// Reads in an Manifest.xml in the given directory and returns it's language specific description. Defaults to an description without any attributes.
+        /// </summary>
+        /// <param name="strDirectory"></param>
+        /// <returns>A description as String, adds a comment if it defaulted to a node without language attribute</returns>
+        public string GetManifestData(string strDirectory)
+        {
+            string strLanguage = GlobalOptions.Language;
+            XmlDocument xmlObjManifest = new XmlDocument();
```
DelnarErsike, line 1868: Have you profiled this? I imagine you'd ideally want to only create and load in the `XmlDocument` once, either at the first request for the manifest or when the Info itself is created/loaded. Constantly re-creating the manifest text every single time a description is requested is not ideal.

Chummer/Classes/clsXmlManager.cs
```diff
@@ -1857,6 +1857,69 @@ public bool Equals(CustomDataDirectoryInfo other)
...
+            string strFullDirectory = strDirectory + "\\manifest.xml";
```
DelnarErsike, line 1869: Rather than using escaped characters, it's probably better to use `Path`'s functions. Should also help you avoid the need for `@` flags everywhere, too.

Chummer/Classes/clsXmlManager.cs
```diff
@@ -1857,6 +1857,69 @@ public bool Equals(CustomDataDirectoryInfo other)
...
+            if (xmlNodeDescription == null)
+                return LanguageManager.GetString("Tooltip_CharacterOptions_ManifestDescriptionMissing");
+
+            string strDescription = xmlNodeDescription.InnerText;
+
+            //Add that the correct language could not be found
+            if (blnDefaultedToEng)
+                strDescription =
+                    LanguageManager.GetString("Tooltip_CharacterOptions_LanguageSpecificManifestMissing") +
+                    "\n" + "\n" + strDescription;
```
DelnarErsike, line 1916: These should be `Environment.NewLine`s, not `\n`s.

Chummer/Classes/clsXmlManager.cs
```diff
@@ -1857,6 +1857,69 @@ public bool Equals(CustomDataDirectoryInfo other)
...
+        public string GetManifestData(string strDirectory)
+        {
+            string strLanguage = GlobalOptions.Language;
+            XmlDocument xmlObjManifest = new XmlDocument();
+            string strFullDirectory = strDirectory + "\\manifest.xml";
+            bool blnDefaultedToEng = false;
+
+            //This will come up very often, so check this here instead of handling the exception. It also catches all exceptions regarding permission and bad strings(Paths)
+            if(!File.Exists(@strFullDirectory))
+                return LanguageManager.GetString("Tooltip_CharacterOptions_ManifestMissing");
```
DelnarErsike, line 1874: It might be annoying to have "manifest missing" pop up as tooltips for most stuff. Maybe it would be better to have tooltips be blank for stuff without any manifest file at all, at least if you stay with the current setup.

Chummer/frmCharacterOptions.cs
```diff
@@ -740,6 +740,7 @@ private void PopulateCustomDataDirectoryTreeView()
                     if (objCustomDataDirectory.Item1 is CustomDataDirectoryInfo objInfo)
                     {
                         objNode.Text = objInfo.Name;
+                        objNode.ToolTipText = objInfo.GetManifestData(objInfo.Path);
```
DelnarErsike, line 743: This particular arrangement works right now because description data is all we'd be using. However, the moment stuff like dependency data or author data or version data would also be used, these tooltips are probably going to get crowded and/or hard to follow with your eyes. Maybe it would be better to have this data appear in a dedicated section within this tab of `frmCharacterOptions` instead of as tooltips. If we do that, we could then also use the tooltips for other stuff, e.g. explanations as to why an enabled custom ruleset might be highlighted a certain color when it has an unmet dependency.

DelnarErsike (commented):
It's working as-is, but I'd still have a look at some of my comments to see if you want to make any significant overhauls before we merge.

Chummer/Classes/clsXmlManager.cs
JoschiGrey: It's called once, for every custom dataset, when opening the Character Options.
I could implement a `Description` property for the `ManifestInfo` class and store the tooltip there.
Then a simple `if`, would check if the property was already set.

Chummer/frmCharacterOptions.cs
JoschiGrey: I would keep this for now, because it is the simplest solution at the moment.
We have to rework this later though.

Something akin to how spell information are displayed could work well. So a layout panel of some sort, that shows the description of the selected `treNode`.

Chummer/Classes/clsXmlManager.cs
JoschiGrey: For me it was fine, but I can see how this is annoying to other users.
Gonna take this out and just return from the method.

Chummer/customdata/manifest.xsd
JoschiGrey: Chummer could calculate the hash and add it to the `.xml` as a `hash /` node the first time a custom data directory is loaded and no hash info is present. I don't know how resource hungry calculating hashes is, we could just calculate it each time. Then the user won't have to delete the hash each time he changed his file.

If a user now wants to create it's own dependency, they would just need to copy out the hash from one file into the next.
Copying over the GUID and Version or the auto generated hash, are probably equivalent in complexity. 

But we should then define dependencies with `<name />` and `<hash />`. The name is just needed for human readability. Maybe including the `<version />` for this purpose as well would be a good idea.

Chummer/customdata/German Data Changes/manifest.xml
JoschiGrey: Good idea. Marking the main author with an attribute and defaulting to the first one in the list would be the way to go.

Chummer/Classes/clsXmlManager.cs
JoschiGrey: I just remembered why I didn't go with a property in the first place.
Calling `GetManifestData()` in the constructor, leads to an `standard text style has to be set before win32 object can be created` exception. After that I just scraped the idea all together.

That said I will just not do it in the constructor of `CustomDataDirectoryInfo`, but in the load of the `frmCharacterOptions` and just check if the property is set already.

Chummer/customdata/manifest.xsd
DelnarErsike: Creating a SHA512 hash/checksum from the contents of files should take milliseconds unless the files in total are larger than about 100 MB. The difficulty is more from the setup of needing to launch Chummer in order get the hash in the first place. Even if it's done automatically, you'd still have to launch Chummer each time you changed something to get an updated hash for your custom ruleset.

JoschiGrey: With hashes wo won't get around that.
We could throw in some additional work to make it more intuitive, to start up chummer for doing stuff like this.

For example a way do create dependencies in chummer.

- Open character files.
- Select a Custom Data.
- There is a add dependencies button.
- You select another directory.
- Chummer does all the background of adding the hash to the correct file.

If we not only use a hash as identifier, we could also update the hash in all other files, that are identified with name and (older) version. That where dependend upon the chabged file.
This would  make updating a data file really easy. As long as you are sure, that you didn't disrupt any dependencies.

JoschiGrey:
I now worked in your initial suggestions and this could be merged now to serve as an early implementation.

I could not find why exactly the exception occurs I mentioned earlier, when calling a method in the constructor, so I took the easy way out for now.

What needs to be done now or after the merge is to create the `manifest.xml` for all our Custom Data directories, that are shipped with Chummer itself. I have no clue, what a few of those are doing or if they are necessary at all.

For later down the line, we need to work out, what and how we want to do next.

Current TODO from our previous discussion:

- Work out and implement a better way of displaying the information of the `Manifest.xml` in chummer. Tooltips won't do if we want to show Author etc.
- Create some form of dependency system, either with (hash based or GUID+Version)?
- Maybe add a flag if a file is downward / upward compatible in the `manifest`?

JoschiGrey:
I moved my work to another Branch, because this was on my master.
It can be found her: https://github.com/JoschiGrey/chummer5a/tree/CDManifest-Expansion
I will open this as an PR as soon as it is a bit more fleshed out.

PR #4447, opened 2021-05-18
Now factors wound modifiers into multiple attribute checks

JoschiGrey:


Chummer/Classes/clsCharacter.cs
```diff
@@ -10898,7 +10898,7 @@ public int CurrentCounterspellingDice
             }
         }
         #region Dodge
-        public int Dodge => REA.TotalValue + INT.TotalValue + TotalBonusDodgeRating;
+        public int Dodge => REA.TotalValue + INT.TotalValue + TotalBonusDodgeRating + WoundModifier;
```
DelnarErsike, line 10901: Could you update the dependency graph as well so that `Dodge` gets its `PropertyChanged` fired whenever `WoundModifier` changes?

DelnarErsike (changes requested):

Chummer/Classes/clsCharacter.cs
JoschiGrey: Can do, but it already worked like this without any problems, for some reason and then forgot about those.

JoschiGrey:
Also fixed the omissions in Judge Intentions, Lift and Carry, Memory and Composure tests.

DelnarErsike (approved):

PR #4452, opened 2021-05-19
Diceroller bug 4451

### LINKED ISSUE #4451: [WIB] Dicetool crash with re-roll and rule of 6
### Environment
Crash ID: 5d722158-7d39-4089-9fdb-a3bffbcb5c3c
Chummer Version: 5.214.0.0
Environment: Windows 10 Home
Runtime: 4.0.30319.42000
Option upload logs set: OnlyMetric
Installation ID: 95bbd100-b9f5-4a33-8c1e-ec2c5db57fb7

### Expected behaviour
I should be able to re-roll dice also if i had more hits than i had dice (due to the rule of 6)

### Actual behaviour
Chummer crashes after several re-rolls

### Steps to reproduce the behaviour
Use a small number of dice in the diceroller
Enable the rule of 6
Roll the dice
if you get more hits than you have dice:
Re-Roll and Chummer will crash
frmDiceRoller.cs at this line:
`List<int> lstRandom = new List<int>(nudDice.ValueAsInt - intHitCount);`

Phil5555:
Fixes #4451, see decription on how to reproduce there.

Solution: Calculate the dicepool to re-roll by counting how many non-hits have been removed from the result list.


DelnarErsike:
Good catch, though normally this should not occur because you're only allowed to use Edge on a roll once, so re-rolling failures is mutually exclusive with Push the Limit's exploding 6's.

PR #4453, opened 2021-05-19
Colored diceroller tool

Phil5555:
Yet another improvement that brings more colors to Chummer.
This PR changes the dicetool and shows hits in green and glichtes in red.


Chummer/Classes/ColorManager.cs
```diff
@@ -224,6 +224,22 @@ private static Color TransformToDarkModeValidVersion(Color objColor)
...
 
+        public static Color DieGlitchFore => IsLightMode ? DieGlitchForeLight : DieGlitchForeDark;
+        public static Color DieGlitchBackground => IsLightMode ? DieGlitchBackgroundLight : DieGlitchBackgroundDark;
+        public static Color DieHitFore => IsLightMode ? DieHitForeLight : DieHitForeDark;
+        public static Color DieHitBackground => IsLightMode ? DieHitBackgroundLight : DieHitBackgroundDark;
+
+        private static Color DieGlitchForeLight => Color.White;
+        private static Color DieGlitchBackgroundLight => Color.DarkRed;
+        private static Color DieHitForeLight => Color.Black;
+        private static Color DieHitBackgroundLight => Color.LightGreen;
```
DelnarErsike, line 235: Bad choice of colors because of red-green colorblindness. There's a reason why the two colors used in the Karma/Nuyen charts are red and blue.

At the same time, does adding color information really help with legibility at a glance? The form already reports how many hits you have and whether or not you glitched. I don't see how coloring the background of individual 1, 5, and 6 entries (with 2's and 4's as well depending on the selected options) in the list is helpful TBH.

Chummer/Classes/ColorManager.cs
```diff
@@ -224,6 +224,22 @@ private static Color TransformToDarkModeValidVersion(Color objColor)
...
+
+        private static Color DieGlitchForeDark => GenerateDarkModeColor(Color.White);
+        private static Color DieGlitchBackgroundDark => GenerateDarkModeColor(Color.DarkRed);
+        private static Color DieHitForeDark => GenerateDarkModeColor(Color.Black);
+        private static Color DieHitBackgroundDark => GenerateDarkModeColor(Color.LightGreen);
```
DelnarErsike, line 240: If you want to cut down on code duplication, these should probably refer to the Light mode colors through their properties instead of directly to the color to which you set those properites.

Chummer/Utilities/frmDiceRoller.Designer.cs
```diff
@@ -157,10 +158,9 @@ private void InitializeComponent()
             this.tlpMain.AutoSize = true;
             this.tlpMain.AutoSizeMode = System.Windows.Forms.AutoSizeMode.GrowAndShrink;
             this.tlpMain.ColumnCount = 3;
-            this.tlpMain.ColumnStyles.Add(new System.Windows.Forms.ColumnStyle());
+            this.tlpMain.ColumnStyles.Add(new System.Windows.Forms.ColumnStyle(System.Windows.Forms.SizeType.Absolute, 80F));
```
DelnarErsike, line 161: Absolute-width columns? Really?

Chummer/Utilities/frmDiceRoller.cs
```diff
@@ -101,22 +103,40 @@ private void cmdRollDice_Click(object sender, EventArgs e)
                     if (chkCinematicGameplay.Checked)
                         intTarget = 4;
 
-                    if (intResult >= intTarget)
+                    if (intResult >= intTarget) {
```
DelnarErsike, line 106: Try to pay attention to the curly bracket standards already established in the rest of the code's style.

Chummer/Utilities/frmDiceRoller.cs
```diff
@@ -101,22 +103,40 @@ private void cmdRollDice_Click(object sender, EventArgs e)
...
                         intHitCount++;
-                    if (intResult <= intGlitchMin)
+                        lviCur.ForeColor = ColorManager.DieHitFore;
+                        lviCur.BackColor = ColorManager.DieHitBackground;
+                    }
+                    if (intResult <= intGlitchMin) {
```
DelnarErsike, line 111: Try to pay attention to the curly bracket standards already established in the rest of the code's style.

Chummer/Utilities/frmDiceRoller.cs
```diff
@@ -285,14 +347,18 @@ private void cmdReroll_Click(object sender, EventArgs e)
                         && (nudDice.ValueAsInt & 1) == 0)))
             {
                 int intBubbleDieResult = GlobalOptions.RandomGenerator.NextD6ModuloBiasRemoved();
-                _lstResults.Add(new ListItem(intBubbleDieResult.ToString(GlobalOptions.InvariantCultureInfo),
-                    LanguageManager.GetString("String_BubbleDie") + strSpace + '(' + intBubbleDieResult.ToString(GlobalOptions.CultureInfo) + ')'));
+                ListViewItem lviCur=new ListViewItem(intBubbleDieResult.ToString(GlobalOptions.InvariantCultureInfo) + strSpace +
+                    LanguageManager.GetString("String_BubbleDie") + strSpace + '(' + intBubbleDieResult.ToString(GlobalOptions.CultureInfo) + ')');
```
DelnarErsike, line 351: Why does this `ListViewItem` have only one argument in its constructor while the others have two? Seems unsafe.

Chummer/Utilities/frmDiceRoller.cs
```diff
@@ -101,22 +103,40 @@ private void cmdRollDice_Click(object sender, EventArgs e)
                     if (chkCinematicGameplay.Checked)
                         intTarget = 4;
 
-                    if (intResult >= intTarget)
+                    if (intResult >= intTarget) {
                         intHitCount++;
-                    if (intResult <= intGlitchMin)
+                        lviCur.ForeColor = ColorManager.DieHitFore;
+                        lviCur.BackColor = ColorManager.DieHitBackground;
```
DelnarErsike, line 109: Make sure you incorporate hooks into the `ColorManager`'s recoloring method so that `ListViewItem` colors are properly updated if/when the user switches between Light and Dark modes while the program is open (e.g. they have it set to Automatic in Chummer and then change their preferred mode in Windows' Settings).

DelnarErsike (changes requested):
I don't really see how this improves things, but if it helps you, it's a good start.
Definitely still needs some more work though. Read my other comments.

Chummer/Classes/ColorManager.cs
JoschiGrey: I personally don't use the dice tool at all, but played around with it after yesterdays Issue that referenced it.
Some coloring would have been nice, because the list was the first thing I looked at after rolling some dice.

Chummer/Utilities/frmDiceRoller.cs
Phil5555: The frmDiceRoller.cs is currently not updated when changing the light/dark mode. Even not when doing it via Windows Settings. This makes it hard to test and I would tend to exclude this into a separate feature request.

Current workaround is to re-open the DiceTool, which will anyway remove the previous results.

Chummer/Utilities/frmDiceRoller.cs
Phil5555: ListViewItems do not have a value any longer. I've cleaned up the second parameter on the other ones. I'm not entirly sure if the TryParse statement still works for the GlobalOptions.CultureInfo format. But we are not dealing with floating point numbers anyway.

Chummer/Classes/ColorManager.cs
Phil5555: I tried to apply the Red/Blue and DarkRed/Blue colors. Finally I now think that Gray/Green makes more sense, but I can change it once more if you like.
Regarding the question if it is useful: It at least allows to identify relevant parts more quickly.

DelnarErsike (approved):

PR #4462, opened 2021-05-26
Custom Data Manifests

JoschiGrey:
This pull request implements support for `manifest.xml` files for custom data folders.

Those manifest files can store information about the directory:

- The author (+ if it is a main author)
- The Version
- Descriptions in multiple languages
    - Descriptions default to the `en-us` description, if no fitting was present.
- Other directories this is dependent upon or that can't be with this one.
   - Dependencies and exclusivities are identified by GUID and Version.
   - Dependencies can set it's version to be a minimum Version, it then accepts all directories with this or an higher version.
   - Exclusivities can ignore version all together.
   - If a dependend directory is missin (not activated) or an exclusive one is present, it will be shown by the node color and a tooltip displays which one should be activated or deactivated.
- All information are displayed in the Character Options Window
image


Chummer/Classes/clsCustomDataDirectoryInfo.cs
```diff
@@ -0,0 +1,475 @@
...
+                if (isMinimumVersion)
+                {
+                    if (objCharacterOptions.EnabledCustomDataDirectoryInfos.All(loadedFile =>
+                        depGuid != loadedFile.Guid && depVersion != loadedFile.Version))
+                        missingDependencyList.Add(string.Format(depName + LanguageManager.GetString("AddMinimumVersion"), depVersion));
+                }
+                else
+                {
+                    if (objCharacterOptions.EnabledCustomDataDirectoryInfos.All(loadedFile =>
+                        depGuid != loadedFile.Guid && depVersion >= loadedFile.Version))
```
DelnarErsike, line 280: The version conditionals look weird to me here. Right now, `isMinimumVersion == true` will show an error if `depVersion != loadedFile.Version`, while `isMinimumVersion == false` will show one if `depVersion >= loadedFile.Version`. To me, it seems like what should happen is `isMinimumVersion == true` should show an error if `depVersion < loadedFile.Version`, and `isMinimumVersion == false` should show one if `depVersion != loadedFile.Version`.

Chummer/Classes/clsCustomDataDirectoryInfo.cs
```diff
@@ -0,0 +1,475 @@
...
+        {
+
+            var existingProhibitedDirectory = new List<string>();
+
+            foreach (var (forbiddenGuid, forbiddenName, forbiddenVersion, ignoreVersion) in ExclusivitiesList)
+            {
+                if (!ignoreVersion)
+                {
+                    if(objCharacterOptions.EnabledCustomDataDirectoryInfos.Any(enabledDirectory =>
+                        enabledDirectory.Guid == forbiddenGuid && enabledDirectory.Version == forbiddenVersion))
```
DelnarErsike, line 304: I kind of feel like equality for the version here wouldn't work practically. If version is important for exclusivity information, I would imagine there could be cases where you'd want to forbid any version older than a specific one, any version newer than a specific one, or both. Maybe some sort of wildcard handling might be better, and also work better with dependency checks, where you also could have similar setups (wanting to be dependent on a specific range of versions.

Chummer/Classes/clsCustomDataDirectoryInfo.cs
```diff
@@ -0,0 +1,475 @@
...
+
+        /// <summary>
+        /// A list of all exclusivities each formatted as Tuple(int guid, str name, int version, bool ignoreVersion).
+        /// </summary>
+        public IReadOnlyList<Tuple<Guid, string, int, bool>> ExclusivitiesList => _lstExclusivities;
+
+        /// <summary>
+        /// A Dictionary containing all Descriptions, which uses the language code as key
+        /// </summary>
+        public IDictionary<string, string> DescriptionDictionary { get; } = new Dictionary<string, string>();
```
DelnarErsike, line 384: Maybe this should be `IReadOnlyDictionary`.

Chummer/Classes/clsCustomDataDirectoryInfo.cs
```diff
@@ -0,0 +1,475 @@
...
+
+        /// <summary>
+        /// The description to display
+        /// </summary>
+        public string DisplayDescription { get; private set; }
+
+        /// <summary>
+        /// A Dictionary, that uses the author name as key and provides a bool if he is the main author
+        /// </summary>
+        public IDictionary<string, bool> AuthorDictionary { get; } = new Dictionary<string, bool>();
```
DelnarErsike, line 394: Maybe this should be `IReadOnlyDictionary`.

Chummer/Classes/clsCustomDataDirectoryInfo.cs
```diff
@@ -0,0 +1,475 @@
...
+
+        public override int GetHashCode()
+        {
+            return new { Name, Path }.GetHashCode();
+        }
+
+        public bool Equals(CustomDataDirectoryInfo other)
+        {
+            return other != null && Name == other.Name && Path == other.Path;
+        }
```
DelnarErsike, line 435: Are you sure these still work? These methods work if a `CustomDataDirectoryInfo` just contains a `Name` and a `Path`, but clearly, you've added a lot more stuff, so these methods should be updated to reflect them.

DelnarErsike (changes requested):
See my individual comments. Some of them are questions or suggestions, but other ones (e.g. the way you show dependency and exclusivity information by adding and removing controls that you create in the middle of the method) definitely need to be addressed before I'd feel comfortable merging this.

Chummer/Classes/clsCustomDataDirectoryInfo.cs
```diff
@@ -0,0 +1,475 @@
...
+{
+    public class CustomDataDirectoryInfo : IComparable, IEquatable<CustomDataDirectoryInfo>, IComparable<CustomDataDirectoryInfo>
+    {
+        private bool _blnCreated;
+        private int _version;
+
+        //Tuple<int Hash, string Name, int Version, bool isMinimumVersion>
+        private readonly List<Tuple<Guid, string, int, bool>> _lstDependencies= new List<Tuple<Guid, string, int, bool>>();
+        //Tuple<int Hash, string Name, int Version, bool ignoreVersion>
+        private readonly List<Tuple<Guid, string, int, bool>> _lstExclusivities = new List<Tuple<Guid, string, int, bool>>();
```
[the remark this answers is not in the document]
JoschiGrey: That's an interesting idea and those tuples are honestly really hard to read.
A struct would certainly be better for readability, I'm certainly gonna take a look at that.

The contained strings and other values are probably bigger than the suggested maximum sizes for structs, but _maybe_ sacrificing some performance for readability and maintainability could worth.

Chummer/Classes/clsCustomDataDirectoryInfo.cs
JoschiGrey: There are two versions of this that I'm thinking about right now.

1. First would be to implement some handling for operators.
What I mean by that is, that one could enter `<version>{biggerthen}</version>`, `<version>{equal}</version>`, `<version>{smallerthen}</version>`. I'm not sure how this can be done, but for that I can take a look at how it's already done at other places.

2. The other one would be just having both a <minimumversion> and <maximumversion> node.

Chummer/Classes/clsCustomDataDirectoryInfo.cs
JoschiGrey: The most comparisons should still work, because path is unique to each `CustomDataDirectoryInfo`, but I will take a look and update them if necessary.

Chummer/Program.cs
```diff
@@ -68,6 +68,11 @@ static void Main()
             //for some fun try out this command line parameter: chummer://plugin:SINners:Load:5ff55b9d-7d1c-4067-a2f5-774127346f4e
             PageViewTelemetry pvt = null;
             var startTime = DateTimeOffset.UtcNow;
+
+            //Both of these need to be set, before any windows are created. 
+            Application.SetCompatibleTextRenderingDefault(false);
+            Application.SetUnhandledExceptionMode(UnhandledExceptionMode.ThrowException);
+
```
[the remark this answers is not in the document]
JoschiGrey: I was really unsure about this and tried to do it some other way, because I was unsure of what problems this could entail.

But I did it in the end, because if I don't call them earlier I always run into `InvalidOperationException`
"You can only call this method before the first window is created by your Windows Forms application.". As soon as I call `Create()` in the constructor of `CustomDataDirectoryInfo`.

I couldn't find why that exception occurs, because AFAICT I don't create any windows anywhere. My original workaround was to call create in every `get` accessor and check before that if the object was already created. But that felt incredibly wrong.
Some brief testing right now showed, that the error doesn't occur as long as they are before 
`SetProcessDPI(GlobalOptions.DpiScalingMethodSetting);`, so they can be in the global mutex.

Chummer/frmCharacterOptions.cs
```diff
@@ -765,11 +822,43 @@ private void PopulateCustomDataDirectoryTreeView()
...
+
+            if (objOldSelected != null)
                 treCustomDataDirectories.SelectedNode = treCustomDataDirectories.FindNodeByTag(objOldSelected);
+            treCustomDataDirectories.ShowNodeToolTips = true;
             treCustomDataDirectories.EndUpdate();
         }
 
+        private void CheckForDependenciesAndExclusivities()
+        {
+            foreach (TreeNode directoryNode in treCustomDataDirectories.Nodes)
```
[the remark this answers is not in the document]
JoschiGrey: The real check for dependencies and exclusivities is only done for nodes which are `checked`.

I did some basic diagnostics a while back and AFAICT it isn't really noticeable in regards to performance, even with a bunch of active directories.

A version in which it is only checked for recently changed once has the problem, that it doesn't update already active directories.
As an example: If I activate Directory "1", that is dependent on "2" it would show up as red.
Now I activate "2", but "1" would still be shown as missing something.

The object of `CustomDataDirectoryInfo` "2" doesn't know if some other directory may be dependent upon it, so we need to check every active dataset.

Chummer/Classes/clsCustomDataDirectoryInfo.cs
ArchonMegalon: If you want to implement it in a sane way (and are willing to put some work into it), the guys from nuget had to come up with a robust way to handle all this version dependencies (probably in a much more clever way than we could in years).

So - if you are "inventing" a syntax I would suggest to take a look at their "Version ranges" syntax for some ideas.

https://docs.microsoft.com/en-us/nuget/concepts/package-versioning

Chummer/Classes/clsCustomDataDirectoryInfo.cs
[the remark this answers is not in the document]
ArchonMegalon: I'm a fan of classes just because I often started out with a struct just to change it to a class later, when I had to maintain the codebase and add some logic, checks of constructor-magic to it. 

My "gut"-feeling would be: Just make it a List<MyDependencyObject> and put everything you need in properties.

Chummer/Classes/clsCustomDataDirectoryInfo.cs
ArchonMegalon: This looks promising! Thank you @JoschiGrey and @DelnarErsike for this!

ArchonMegalon:
I need a bit of github-expertise here: how did you apply the "change requested" status here?

ArchonMegalon (changes requested):
I found it! :)

Chummer/Classes/clsCustomDataDirectoryInfo.cs
JoschiGrey: That's a nice syntax for sure, if I find a nice way to implement and handle it I will just use that.

But maybe just using inclusive min and max versions without any syntax is enough and way easier to use.

Chummer/Classes/clsCustomDataDirectoryInfo.cs
JoschiGrey: Looking back at them, they are truly a mess right now and will be completely replaced with the new version handling we are discussing down below.

JoschiGrey:
Changes:

- Replaced dependency quad tuple list by `class DirectoryDependency`
- Replaced a bunch of wonky string creation with `StringBuilder`s
- Changed `Version` from `int` to `decimal`
- Removed some unnecessary method calls
- Moved all `private` create helper methods into the constructor as local methods
- Reverted changes to the global mutex
- Moved all `LanguageManager.GetString` usages into an `LazyCreate()` method, which is called as soon as `frmCharacterOptions` is opened. This allows the previous change to the mutex to be reverted. See my comments in `clsCustomDataDirectoryInfo.LazyCreate()` for explanations.
- Replaced usages of `TryGetField()` with their respective `TryGetTYPEQuickly` variants
- Changed the version handling to use `<minversion>` and `<maxversion>`, which are handled as an inclusive version range. They default to `dec.MinValue` and `dec.MaxValue` respectively.
- Updated `manifest.xsd`.
- Made a bunch of fields and properties `readonly `or `private set`
- Updated `frmCharacterOptions` to replace a `Label.Text` with an formatted string instead of creating and deleting labels to display dependencies.

Chummer/frmCharacterOptions.cs
```diff
@@ -734,6 +782,7 @@ private void PopulateCustomDataDirectoryTreeView()
                     if (objCustomDataDirectory.Item1 is CustomDataDirectoryInfo objInfo)
                     {
                         objNode.Text = objInfo.Name;
+                        objInfo.LazyCreate();
```
DelnarErsike, line 785: Shouldn't this be the other way around?

Chummer/frmCharacterOptions.cs
```diff
@@ -675,6 +677,52 @@ private void cboPriorityTable_SelectedIndexChanged(object sender, EventArgs e)
...
+            lblDirectoryName.Text = objSelected.Name;
+
+            if (objSelected.DependenciesList.Count > 0)
+            {
+                StringBuilder sb = new StringBuilder();
+
+                foreach (var dependency in objSelected.DependenciesList)
+                {
+                    sb.Append(dependency.DisplayName);
+                    sb.Append(Environment.NewLine);
```
DelnarErsike, line 698: This can be shortened to just one `sb.AppendLine(dependency.DisplayName)`.

Chummer/frmCharacterOptions.cs
```diff
@@ -675,6 +677,52 @@ private void cboPriorityTable_SelectedIndexChanged(object sender, EventArgs e)
...
+
+            if (objSelected.ExclusivitiesList.Count > 0)
+            {
+                //We only need a Stringbuilder if we got anything
+                StringBuilder sb = new StringBuilder();
+
+                foreach (var exclusivity in objSelected.ExclusivitiesList)
+                {
+                    sb.Append(exclusivity.DisplayName);
+                    sb.Append(Environment.NewLine);
```
DelnarErsike, line 716: This can be shortened to just one `sb.AppendLine(exclusivity.DisplayName)`.

Chummer/frmCharacterOptions.cs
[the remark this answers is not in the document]
DelnarErsike: I was thinking more in terms of setting up some kind of graph when all the `CustomDataDirectoryInfo`s are first created that links up all the different relations they have. Then when you change the enabled status or load order of one info, you just need to query all the infos connected to it in that graph. However, I admit that's probably a bit much for now and unnecessary to get the first working versions of these manifests usable by the public, so you can probably file that away as a TODO somewhere near this code and move on.

Chummer/frmCharacterOptions.cs
```diff
@@ -765,11 +814,50 @@ private void PopulateCustomDataDirectoryTreeView()
...
+                        missingDirectories = customDataDirectory.CheckDependency(_objCharacterOptions);
+
+                    string prohibitedDirectories = string.Empty;
+                    if (customDataDirectory.ExclusivitiesList.Count>0)
+                        prohibitedDirectories = customDataDirectory.CheckExclusivity(_objCharacterOptions);
+
+                    if (string.IsNullOrEmpty(missingDirectories) && string.IsNullOrEmpty(prohibitedDirectories))
+                    {
+                        directoryNode.ToolTipText = string.Empty;
+                        directoryNode.ForeColor = Color.Empty;
```
DelnarErsike, line 843: Should this be `Color.Empty`? What color is that in practice, and how does it interact with `ColorManager` and dark mode?

Chummer/frmCharacterOptions.cs
```diff
@@ -765,11 +814,50 @@ private void PopulateCustomDataDirectoryTreeView()
...
+                    string tooltip =
+                        CustomDataDirectoryInfo.BuildExclusivityDependencyString(missingDirectories, prohibitedDirectories);
+
+                    directoryNode.ToolTipText = tooltip;
+                    directoryNode.ForeColor = ColorManager.ErrorColor;
+                }
+                else
+                {
+                    directoryNode.ToolTipText = string.Empty;
+                    directoryNode.ForeColor = Color.Empty;
```
DelnarErsike, line 856: Should this be `Color.Empty`? What color is that in practice, and how does it interact with `ColorManager` and dark mode?

DelnarErsike (changes requested):
You're getting close, but there are still a few things that need cleaning up, mainly involving (mis)use of `StringBuilder` and doing some O(n^2) things that probably shouldn't be O(n^2).

Chummer/frmCharacterOptions.cs
JoschiGrey: You mean the lazy create should be before the other to assignments?
That is not really necessary right now, because `Name` and `Path` are normally set in the constructor.

But it may be clearer or better to read for others to move it up a bit.

Chummer/frmCharacterOptions.cs
JoschiGrey: I will chek how this interacts with the dark mode, but in theory this should just remove the additional color from the node.

JoschiGrey:
I resolved everything that I finished now.
Notable exceptions:
- The GetHashCode and Equals
- The ColorCode stuff

The new pattern for dependency checks should roughly come out to O(log(n) + n + 2) in the worst case but should be O(n + 3) in normal use cases. (As long as I understood O notation correctly)

Gonna do the rest in the next few days

ArchonMegalon (approved):
I opt out of this branch, because I don't have the time right now to review it. But following the discussion I know this in good hands! :)

Chummer/frmCharacterOptions.cs
JoschiGrey: So this interacts well with the dark mode. It resets it to the default color with `Color.Empty`

Chummer/Classes/clsCustomDataDirectoryInfo.cs
```diff
@@ -0,0 +1,611 @@
...
+        {
+            if (obj is CustomDataDirectoryInfo objOther)
+                return Equals(objOther);
+            return false;
+        }
+
+        public override int GetHashCode()
+        {
+            //Path and Guid should already be enough because they are both unique, but just to be sure.
+            return new { Name, Path, Guid, Version, DependenciesList, ExclusivitiesList }.GetHashCode();
```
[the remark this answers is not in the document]
JoschiGrey: So there is `EnumeralExtension.GetEnsembleHashCode<T>()` which produces a hash code based on the content of the `IEnumerable`. I can't however just include those in the `GetHashCode()`.

Calculating them beforehand and then including the pointer to those local variables like this:

``` c#
        public override int GetHashCode()
        {
            var dependencyHash = DependenciesList.GetEnsembleHashCode();
            var exclusivityHash = ExclusivitiesList.GetEnsembleHashCode();
            //Path and Guid should already be enough because they are both unique, but just to be sure.
            return new { Name, Path, Guid, Version, exclusivityHash, dependencyHash }.GetHashCode();
        }
```
would work.

But my problem with this is, that I'm not sure if I understood everything.
What I think this way of doing it would do compared to the previous implementation is as follows.

`GetHashCode` takes in all the stuff given in the brackets {} and calculates a Hash. But it can't use a `IEnumerable`, so there instead of the content the pointer is used to calculate the hash.
If I now give it the `Int` hashes of those lists instead of the lists themselves those integers are used for the Hash calculation and in extension our resulting hash depends on the content of those lists.

DelnarErsike (approved):
I saw a few more things that might need work, but I'm fine doing those myself.

PR #4470, opened 2021-06-01
Not saying Sinners works now... 

ArchonMegalon:
- just that it is less broken than before.

Authentication should work
Handling Groups that are NULL works


ArchonMegalon:
Leave this open for now please. Will have to check if it also works if the client is NOT behind a proxy...

PR #4486, opened 2021-06-09
Seems like I can't add a .git folder. Added a git folder and a pre-bu…

ArchonMegalon:
…ild step, that copies the file.


ArchonMegalon:
Please check, if this works how it should!

DelnarErsike:
`.git` is always purely local, so commit pre-hooks that you set up in there will always be purely local. You cannot push `.git` by design, that folder is always for your local stuff alone. You would need to put that pre-hook directly into the repository server's `.git` directory.

Now reading up on Git hooks for GitHub, GitHub does not allow arbitrary hooks for non-Enterprise repositories; Enterprise repositories are self-hosted, which is why they're allowed there. The only things GitHub allows are interactions with their proprietary webhooks and integrated third-party services. So this approach is definitely not going to work.

ArchonMegalon:
Well... why not? I implemented a client-side hook and it runs locally before even pushing to github. The only thing needed to be done (and that's why it's a pre-build-step in the chummer.project is to actually copy the file into the .git folder.

DelnarErsike:
Read my comment again. The `.git` folder is purely local. You can never commit it directly; or maybe you can, but you never, ever, *ever* should, the same you can technically log in and use Linux with `root` to never have to bother with permissions.

In the case of GitHub that is not GitHub Enterprise, `.git` is completely locked and can only be modified through the limited, official hooks that GitHub provides. GitHub Enterprise involves self-hosting a Git server, which is not really viable for us.

ArchonMegalon:
But I am not commiting the .git folder, but another one called "git" (without the "."). So github should not have a problem with that.

And what I do locally (like copying the content of "git" to ".git" also does not concern github. 

The only thing that happens after the file is copied into the ".git" folder is, that the local git.exe findes the pre-commit script and executes it before the commit (again - no involvement of github here).

ArchonMegalon:
Of course it would be better (policy-wise) to enforce the hook-script on the server. But on the other hand - it's more just a hint for the developer. I see no harm in it running on the client side.

PR #4514, opened 2021-07-23
Implemented TryGetVersionFieldQuickly

### LINKED ISSUE #4513: Chummer5a crashes after clicking on Custom Data folder with Manifest
**Describe the bug**
Clicking on a Custom Ruleset containing a Manifest file results in Chummer5a crash.

**To Reproduce**
Click on Custom Ruleset with a Manifest file.

**Expected behavior**
Program displays content of Manifest file without crashing.

**Screenshots**
[chummer_crash_132714748010769928.txt](https://github.com/chummer5a/chummer5a/files/6866081/chummer_crash_132714748010769928.txt)

Screen recording of select and crash
https://user-images.githubusercontent.com/17562417/126726362-99b116e3-0e41-4371-9bfc-af7ead637f88.MP4

 - OS: Windows 10 Home
 - Version: 5.214.130 (also 129)

JoschiGrey:
fixes #4513 

`TryGetField` interacts badly with `Version.TryParse`, because this will output `null` and not default to `TryGetField`s error default value. This then leads to `NullPointerExceptions`. 

`TryGetVersionFieldQuickly` uses `version.TryParse`. If this outputs null it tries `int.TryParse` to get it as a `Major` for `Version`.

As an example
`<version>1</version>` will evaluate to `1.0` instead of giving out `null`.

### DIFF
--- a/Chummer/Backend/Extensions/XmlNodeExtensions.cs
+++ b/Chummer/Backend/Extensions/XmlNodeExtensions.cs
@@ -526,6 +526,39 @@ namespace Chummer
         }
 
         /// <summary>
+        /// Like TryGetField for Version, but taking advantage of version.TryParse... boo, no TryParse interface! :(
+        /// If TryParse outputs null it will to try to use int.TryParse and use the output as a major. e.g. 1 becomes 1.0
+        /// </summary>
+        /// <param name="node"></param>
+        /// <param name="field"></param>
+        /// <param name="read"></param>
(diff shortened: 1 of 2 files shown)


DelnarErsike:
Unnecessary because a better version was added in commit 57ee7a9229eab7fedba45aca5dba61e85e58c036. Having 1.0 be the default value for `Version` objects is not great because it can and will trip up if people define their custom data manifests with versions like 0.1 or 0.2.

PR #4550, opened 2021-08-11
Correct German translation for 'Touch'

Lochabar:
German sheets xz.Language file has incorrect translation of 'Touch' - should be 'Bouch" (instead of 'Berührung')


DelnarErsike:
I'm pretty sure this wasn't the problem, and the real problem was already fixed in commit 8c199d44bed37d07d6e970b92be71e3bdaad0e0b. "Berührung" is correct in German, whose shortened form is "B", the same way that "Touch" is shortened to "T" in English. However, some critter powers had ranges defined as "Touch" instead of "T". The code would see the "T" in "Touch" and translate it to "B" as per the German form, leading to it showing up as "Bouch".

Issue #4599, opened 2021-11-03
Shaman Tuxedo asks for Spell Category instead of Spirit Type


ArchonMegalon:
Cutting Aces Page 143:

RAW: "Each set is designed to affect one specific type of spirit that tradition is capable of conjuring."

It asks for the Spell Category.

larryw24:
Is this a duplicate of Bug #4225?  https://github.com/chummer5a/chummer5a/issues/4225

PR #4600, opened 2021-11-03
Fix for Shaman Tuxedo should ask for of Spirit Type 

ArchonMegalon:
See #4599


ArchonMegalon:
@DelnarErsike  Since you are kind of the one, who knows the most about chummer I just wanted to show you the solution I implemented for "insert_something" into the xpath.

DelnarErsike (changes requested):
I don't like it. The `@` symbol is already used in XSLT/XPath to denote an attribute. Using it as a way macro in character-dependent stuff seems like it's asking for things to go wrong later if/when someone decides to add a macro that lines up with certain attributes in one of the data files.

Plus, the way you've implemented the Shaman Tuxedo selection here is going to fall apart for Custom traditions. As clumsy as it might be, I honestly think it's better to just not filter based on the character's current tradition and let people pick any single spirit they want for their Shaman Tuxedo.

ArchonMegalon:
Well, the first part is easy. We can exchange the @ for every other symbol. But the Custom traditions were kind of roaming around in my head when I asked you for your opinion. I don't know enough about the custom amends structure inside chummer. Is there a way to parse these as well? Presenting all the Spirits (and they are the same for a lot of traditions) is kind of... ugly.

DelnarErsike:
Custom traditions aren't in the amends system, they're their own thing inside Chummer where users can just select the attributes and spirit types they want for each spell category and Chummer will accommodate.

There is also the third bit I mentioned which is that someone might want to buy a Shaman Tuxedo for a spirit type that isn't one supported by their tradition. This could be either for RP reasons (e.g. they might be buying it for someone else and want it to show up in their expense logs regardless) or because they're not a magical character and are buying the Shaman Tuxedo for a magical NPC.

chummer5a:
Tuxedos are always tradition-linked so _technically_ wouldn't work with UMT expansion and such. I guess the other trick would be to use two select-texts, one that selects the tradition, the second that selects a spirit.

ArchonMegalon:
Maybe it will work with some form of distinct values:

https://stackoverflow.com/questions/2812122/distinct-in-xpath

will take a look at this next week. As long as Custom Traditions don't introduce totaly new Spirit Types, this should be fine.

ArchonMegalon:
Killing this for now. No fast and free different Xpatz library found.

PR #4610, opened 2021-11-12
Disentangle UI and backend [WIP]

JoschiGrey:
This draft should just show, how I intend on separating UI and backend code and allow for early comments, corrections and suggestions.

The big question for me right now is:
Many of those Methods could just go into the `Character` class, instead of putting them into their own `namespace` and `class`,
but `character` is already a giant unwieldy class and file.

Making it `public sealed partial Character` would allow to split them into other files, but it would decrease readability even further. 

That is why my solution right now is to open a new Namespace `Chummer.Backend.BuySellIncreaseDecreaseMethods` and introduce a few `public static class NAME`. Those should group those methods into a few coherent classes and keep file sizes low for readability.


Chummer/Backend/Buy Sell Increase Decrease Methods/ChangeAttributes.cs
```diff
@@ -0,0 +1,43 @@
...
+            characterObj.ExpenseEntries.AddWithSort(objExpense);
+            characterObj.Karma -= intKarmaCost;
+
+            ExpenseUndo objUndo = new ExpenseUndo();
+            objUndo.CreateKarma(KarmaExpenseType.AddPowerPoint, string.Empty);
+            objExpense.Undo = objUndo;
+
+            characterObj.MysticAdeptPowerPoints += 1;
+
+            frmCareer.IsCharacterUpdateRequested = true;
```
chummer5a, line 39: Static shouldn't be responsible for telling the form that it's dirty. Recommend making the method a boolean and have it perform the character update/dirty state change if it returns true.

JoschiGrey: I already moved it back in a newer version, because I suspected it was the wrong place for that.

But the bool return is a good idea, gonna adopt that.

JoschiGrey:
Would it be a good idea to just implement all those `static` methods as extensions to `Character` instead of `static public` methods with `Character character` as one of it's params?

JoschiGrey:
This is obviously not done, but some regular merges from this branch into master would be great.
The changes should not impact the functionality of Chummer and I would really like to not having to deal with
resolving conflicts while merging that much.

DelnarErsike:
> This is obviously not done, but some regular merges from this branch into master would be great. The changes should not impact the functionality of Chummer and I would really like to not having to deal with resolving conflicts while merging that much.

Then you'll have to copy this over to a branch on Chummer itself so that you can cherry-pick commits in it directly onto `master`. When they're on separate repositories, the only ways to merge changes from one to the other would be to either copy them over manually or to merge pull requests (which merges every commit from one branch into the other).

JoschiGrey:
I don't have the necessary permissions on git to create and maintain a branch on Chummer itself.

DelnarErsike:
If you (or someone else with access to your repo) resolve the merge conflicts, we could try merging it over as-is.

PR #4632, opened 2021-12-08
Added "armorcapacity" to JSON export of armor node

Eliasar:
I'm unsure of the changes in Contributors.cs, so please double check that [in the contributors page here](https://github.com/chummer5a/chummer5a/graphs/contributors).

### My conjecture
The users that were removed did not show up in that page, but I suspect that has to do with their account or possibly removing their forked project


Chummer/Backend/Equipment/Armor.cs
```diff
@@ -856,6 +856,7 @@ public void Print(XmlTextWriter objWriter, CultureInfo objCulture, string strLan
             objWriter.WriteElementString("category", DisplayCategory(strLanguageToPrint));
             objWriter.WriteElementString("category_english", Category);
             objWriter.WriteElementString("armor", DisplayArmorValue);
+            objWriter.WriteElementString("armorcapacity", ArmorCapacity);
```
DelnarErsike, line 859: I think you're printing the wrong `ArmorCapacity`. `ArmorCapacity` is the string that Chummer uses internally as a baseline to calculate `TotalArmorCapacity`, which is probably what you want. However, `TotalArmorCapacity` is also used to calculate strings for the calculated capacity and the amount of capacity remaining, so you might want to export those, too. Also, overall, both `TotalArmorCapacity` and the calculated capacity output strings that are based on the decimal formatting of the current UI language. Because the printed language can be different from the UI language, you'd also probably want to change and/or shift code around to let you get versions of those properties that are formatted in any language you'd want instead of in just the UI language.

DelnarErsike (changes requested):
Please have a look to make sure you're actually printing the quantity you want to print (and also, to make sure that whatever you do end up printing will be formatted with the number formatting of the language used for printing instead of the UI language).

Chummer/Backend/Equipment/Armor.cs
Eliasar: This is great feedback. I've been looking through this code for a bit trying to absorb the difference between the member variables and properties in both the armor and armormod files.

I'll go back to the code and review what it is that I want to print in the JSON

DelnarErsike: Also, side note, but Contributors.cs is not a file you're supposed to edit manually, even if you're fetching info from the page you linked. It's updated automatically via a T4 macro that uses GitHub's API (and a private access token that is set up on the AppVeyor build machine that's used to create official builds).

Eliasar: Thank you for the clarification. I've never used a T4 template before, and didn't want to remove someone's name from contribution.

Is this something that should be updated in the Contributing to Chummer Wiki?

Eliasar: Also, the file was edited by running the template in Visual Studio, only to make mention that I did not edit it manually

Eliasar: `CapacityRemaining` is a decimal, so that can be printed out with the correct decimal place mark. `TotalArmorCapacity` looks like it's only ever a string that looks like an integer. Unless the capacity somehow is a decimal, I believe I can just print `TotalArmorCapacity` as it's basically language agnostic

DelnarErsike:
If you are printing `TotalArmorCapacity` and `CalculatedCapacity`, you should probably change both of those to methods that can take a `CultureInfo` (or probably just an `IFormatProvider`) as an argument. Right now, because they're only used by UI methods, they will always use the `CultureInfo` of the current Chummer UI language, but if you want them to be exported via printing, you need to change the way they're used so that they can use the UI language's number formatting when showing stuff in the UI and printed language's number formatting when exporting data to be printed.

Eliasar:
Is `TotalArmorCapacity` or `CalculatedCapacity` ever not an `int` in `string` form?

I am thinking that I'll create a print version of these, since these properties are used in the UI.

Would there be any issue using a method naming scheme of `PrintX` instead of `DisplayX`? Display could be used for UI and I wanted to not cause confusion while conforming to current naming standards

DelnarErsike:
Both of them can end up having square brackets wrapped around them to denote some sort of internal capacity and/or slashes denoting separate capacity cost vs. capacity stats. Even if there is no current item in the dataset where the difference really matters, remember that any modifications you make should follow the spirit of what the output should be, because even if the extra work is pointless *right now*, it won't necessarily be a few months or a year down the line when suddenly someone complains that their custom armor stuff isn't showing up properly in their exported data.

DelnarErsike:
No longer required due to commit 8198e53255d4736d168961c81165092e2e851367.

Issue #4645, opened 2021-12-18
Locks in Parallelized Enumerators Are Not Releasing Properly


DelnarErsike:
As the title suggests, when any sort of `AsParallel()` method is invoked on one of the locking-based collections in Chummer's codebase, read locks are not released properly when iterators are fetched and operated on. Among other things, this causes hangs when adding cyberware plugins to cyberware, e.g. adding Smartlink to a Basic Cybereyes System.

DelnarErsike:
Re-flagged as Showstopper because it's apparently a problem with Plugins.

DelnarErsike:
See the attached image.
![image](https://user-images.githubusercontent.com/727731/150906616-810a9567-e21a-493f-b744-40df660aa658.png)

DelnarErsike:
Nope, still an issue according to Discord.

DelnarErsike:
Confirmed to no longer be an issue after testing with `AsyncFriendlyReaderWriterLocker`.

PR #4677, opened 2022-01-25
Fix for #4645

ArchonMegalon:
@DelnarErsike please take a look, since I don't really understand the Locking-Mechanic.

Problem is: the Plugin messes with the Collection in the Callback. Now - that is probably wrong in itself, but even IF a Plugin does that, the main-program should not crash. That's why I suggest to cast the callback-enumerator to an array before we aggregate it.

#4645

### DIFF
--- a/Chummer/Backend/Characters/Character.cs
+++ b/Chummer/Backend/Characters/Character.cs
@@ -2697,7 +2697,7 @@ namespace Chummer
 
             if (callOnSaveCallBack)
             {
-                blnErrorFree = DoOnSaveCompleted.Aggregate(blnErrorFree, (current, funcToRun) => funcToRun(this) && current);
+                blnErrorFree = DoOnSaveCompleted.ToArray().Aggregate(blnErrorFree, (current, funcToRun) => funcToRun(this) && current);
             }
             return blnErrorFree;
         }
--- a/Chummer/Forms/EditGlobalSettings.cs
+++ b/Chummer/Forms/EditGlobalSettings.cs
@@ -1522,8 +1522,16 @@ namespace Chummer
                                          : strFile,
                                      xmlSheet.SelectSingleNodeAndCacheExpression("name")?.Value ?? string.Empty));
                 }
-
-                string strOldSelected = cboXSLT.SelectedValue?.ToString() ?? string.Empty;
+                string strOldSelected;
+                try
+                {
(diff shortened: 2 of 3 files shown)


Chummer/Forms/EditGlobalSettings.cs
```diff
@@ -1522,8 +1522,16 @@ private void PopulateXsltList()
...
                                      xmlSheet.SelectSingleNodeAndCacheExpression("name")?.Value ?? string.Empty));
                 }
-
-                string strOldSelected = cboXSLT.SelectedValue?.ToString() ?? string.Empty;
+                string strOldSelected;
+                try
+                {
+                    strOldSelected = cboXSLT.SelectedValue?.ToString() ?? string.Empty;
+                }
+                catch(IndexOutOfRangeException)
```
DelnarErsike, line 1530: Wait, what is this for?

Chummer/Backend/Characters/Character.cs
```diff
@@ -2697,7 +2697,7 @@ public bool Save(string strFileName = "", bool addToMRU = true, bool callOnSaveC
 
             if (callOnSaveCallBack)
             {
-                blnErrorFree = DoOnSaveCompleted.Aggregate(blnErrorFree, (current, funcToRun) => funcToRun(this) && current);
+                blnErrorFree = DoOnSaveCompleted.ToArray().Aggregate(blnErrorFree, (current, funcToRun) => funcToRun(this) && current);
```
DelnarErsike, line 2700: In theory, this will fix the issue of the crash. However, it will mean that any `DoOnSaveCompleted` functions that add extra functions to the end of `DoOnSaveCompleted` wouldn't work. I don't know how important that would be though.

DelnarErsike (commented):
Should fix the crash itself, but I don't know if you'd like the side-effects of it. If you're fine with them, merge away.

Chummer/Forms/EditGlobalSettings.cs
ArchonMegalon: You can remove that if you want. Was in my branch because I had a selectedValue of 111 (out of range) and just wanted the program not to crash because of that. Didn't bother tough, WHY the value was that high.

Chummer/Backend/Characters/Character.cs
ArchonMegalon: I think we can live with that.

PR #4683, opened 2022-01-31
Update to net core 6

ArchonMegalon:


ChummerHub/Startup.cs
```diff
@@ -117,12 +118,23 @@ public void ConfigureServices(IServiceCollection services)
...
             //services.AddSingleton<ITelemetryProcessorFactory>(sp => new SnapshotCollectorTelemetryProcessorFactory(sp));
+            Microsoft.ApplicationInsights.AspNetCore.Extensions.ApplicationInsightsServiceOptions aiOptions
+                = new Microsoft.ApplicationInsights.AspNetCore.Extensions.ApplicationInsightsServiceOptions();
+            // Disables adaptive sampling.
+            aiOptions.EnableAdaptiveSampling = true;
 
-            var tcbuilder = TelemetryConfiguration.Active.TelemetryProcessorChainBuilder;
-            tcbuilder.Use(next => new GroupNotFoundFilter(next));
+            // Disables QuickPulse (Live Metrics stream).
+            aiOptions.EnableQuickPulseMetricStream = true;
```
DelnarErsike, line 127: Should these be `false` instead of `true`?

DelnarErsike (commented):
Looks OK from my end, but I'm not super-familiar with ChummerHub's code.

ArchonMegalon:
Well, it's not that we (I) really need it, but in this case it's actually
the Server that broadcasts the QuickPulse to AzureInsights. That is more
informative than some random Chummer-Installation.

By the way, we still have 94 "Servers" reporting back to AI... :/

PR #4686, opened 2022-02-01
Tweaks to AI, CrashHandler, CrashReporter

ArchonMegalon:


Chummer/Backend/Helpers/Application Insights/CustomTelemetryInitializer.cs
```diff
@@ -27,13 +27,14 @@ namespace Chummer
...
+       
         private static Logger Log { get; } = LogManager.GetCurrentClassLogger();
 
         // Set session data:
         //private static string Hostname =  Dns.GetHostName();
-        private static string Version { get; } = System.Reflection.Assembly.GetExecutingAssembly().GetName().Version.ToString();
+        private static string Version { get; } = typeof(Chummer.Program).Assembly.GetName().Version.ToString();
 
-        private static bool IsMilestone { get; } = System.Reflection.Assembly.GetExecutingAssembly().GetName().Version.Revision == 0;
+        public static bool IsMilestone { get; } = typeof(Chummer.Program).Assembly.GetName().Version.MinorRevision == 0;
```
ArchonMegalon, line 37: @DelnarErsike can you please check? I get "IsMileStone"-Logs with 5.220.2.0 in AI...

DelnarErsike: In the distant past of 2015, Chummer's versions were indeeed something like `0.5.172.5`, meaning the major version was 0, the minor was 5, the build was 172, and the revision was 5. The format has been `5.220.2.0` for a while now though, in which case the major version is 5, the minor verison is 220, the build is 2, and the revision is 0. AFAIK, `MajorRevision` and `MinorRevision` just get the highest and lowest 16 bits of `Revision` respectively; if `Revision` is 0, then `MinorRevision` will be 0, too.

PR #4717, opened 2022-02-26
Adding in the Ontario Arms Sling-Shot from SOTA2080

KoshiirRa:
Noticed there was an outstanding request for it, and I'm wanting to use them on a character myself anyway.

### DIFF
--- a/Chummer/data/gear.xml
+++ b/Chummer/data/gear.xml
@@ -1534,42 +1534,43 @@
     </gear>
 	<gear>
       <id>d71ea9c1-9b5f-428e-b849-7d665eb31c5e</id>
-      <name>Ammo: Sling-Shot Acid Capsules</name>
+      <name>Ammo: Slingshot Acid Capsules</name>
       <category>Ammunition</category>
       <rating>0</rating>
       <source>SOTG</source>
       <page>13</page>
-      <avail>-</avail>
+      <avail>0</avail>
       <cost>100</cost>
       <costfor>10</costfor>
       <weaponbonus>
-        <damagereplace>6P Acid</damagereplace>
+        <damagereplace>6P</damagereplace>
+		<damagetype>Acid</damagetype>
       </weaponbonus>
       <ammoforweapontype>slingshot</ammoforweapontype>
     </gear>
 	<gear>
       <id>e20f021b-1d00-4443-a481-d417114e067e</id>
-      <name>Ammo: Sling-Shot Paint Capsules</name>
+      <name>Ammo: Slingshot Paint Capsules</name>
       <category>Ammunition</category>
       <rating>0</rating>
       <source>SOTG</source>
       <page>13</page>
-      <avail>-</avail>
+      <avail>0</avail>
       <cost>10</cost>
       <costfor>10</costfor>
       <weaponbonus>
-        <damagereplace>0</damagereplace>
+        <damagereplace>0S</damagereplace>
       </weaponbonus>
       <ammoforweapontype>slingshot</ammoforweapontype>
     </gear>
 	<gear>
       <id>5ece577b-7c15-4f2c-9f4b-cd325154cf9c</id>
-      <name>Ammo: Sling-Shot Poison Capsules</name>
+      <name>Ammo: Slingshot Poison Capsules</name>
       <category>Ammunition</category>
       <rating>0</rating>
       <source>SOTG</source>
       <page>13</page>
-      <avail>-</avail>
+      <avail>0</avail>
       <cost>100</cost>
       <costfor>10</costfor>
       <weaponbonus>
@@ -1584,11 +1585,11 @@
       <rating>0</rating>
       <source>RG</source>
       <page>23</page>
-      <avail>-</avail>
+      <avail>0</avail>
       <cost>5</cost>
       <costfor>10</costfor>
       <weaponbonus>
-        <damagereplace>As Substance Within</damagereplace>
+        <damagereplace>As Drug/Toxin</damagereplace>
 		<userange>Light Pistol</userange>
       </weaponbonus>
       <ammoforweapontype>slingshot</ammoforweapontype>
--- a/Chummer/data/gear.xml
+++ b/Chummer/data/gear.xml
@@ -1532,6 +1532,51 @@
       </weaponbonus>
       <ammoforweapontype>firefighting cannons</ammoforweapontype>
     </gear>
+	<gear>
+      <id>d71ea9c1-9b5f-428e-b849-7d665eb31c5e</id>
+      <name>Ammo: Sling-Shot Acid Capsules</name>
+      <category>Ammunition</category>
+      <rating>0</rating>
+      <source>SOTG</source>
+      <page>13</page>
+      <avail>-</avail>
+      <cost>100</cost>
+      <costfor>10</costfor>
+      <weaponbonus>
+        <damage>6P Acid</damage>
+      </weaponbonus>
+      <ammoforweapontype>sling-shot</ammoforweapontype>
+    </gear>
+	<gear>
+      <id>e20f021b-1d00-4443-a481-d417114e067e</id>
+      <name>Ammo: Sling-Shot Paint Capsules</name>
+      <category>Ammunition</category>
+      <rating>0</rating>
+      <source>SOTG</source>
+      <page>13</page>
+      <avail>-</avail>
+      <cost>10</cost>
+      <costfor>10</costfor>
+      <weaponbonus>
+        <damage>0</damage>
+      </weaponbonus>
+      <ammoforweapontype>sling-shot</ammoforweapontype>
+    </gear>
+	<gear>
+      <id>5ece577b-7c15-4f2c-9f4b-cd325154cf9c</id>
+      <name>Ammo: Sling-Shot Poison Capsules</name>
+      <category>Ammunition</category>
+      <rating>0</rating>
+      <source>SOTG</source>
+      <page>13</page>
+      <avail>-</avail>
+      <cost>100</cost>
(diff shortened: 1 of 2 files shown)


chummer5a:
Ontario or Onotari?

KoshiirRa:
Yeah, not sure if its a typo on Pegasus's part or not.  There's no info in the fluff paragraph for the weapon indicating it should be Onatari Arms, so maybe its just a similarly-named company?

KoshiirRa:
Yeah, not sure if its a typo on Peagsus's part or not, but nothing in the weapon description indicates that it should be Onotari.

Chummer/data/weapons.xml
```diff
@@ -14843,6 +14843,45 @@
...
+              <category>Light Machine Guns</category>
+              <category>Medium Machine Guns</category>
+              <category>Heavy Machine Guns</category>
+              <category>Assault Cannons</category>
+              <category>Grenade Launchers</category>
+              <category>Missile Launchers</category>
+            </OR>
+          </AND>
+        </OR>
+      </required>
```
DelnarErsike, line 14883: I am pretty sure this is not going to work the way you think it will. The `required` node checks broad character conditions, not the specific details of a single item.

Chummer/data/gear.xml
```diff
@@ -1531,6 +1531,51 @@
...
+      <name>Ammo: Sling-Shot Acid Capsules</name>
+      <category>Ammunition</category>
+      <rating>0</rating>
+      <source>SOTG</source>
+      <page>13</page>
+      <avail>-</avail>
+      <cost>100</cost>
+      <costfor>10</costfor>
+      <weaponbonus>
+        <damage>6P Acid</damage>
```
DelnarErsike, line 1546: This will not work. The `<damage>` node modifies the existing damage number of a weapon, and the `<damagetype>` node modifies the existing damage type of a weapon (stun or physical). You are probably looking for `<damagereplace>`, which outright replaces the damage code of a weapon.

Chummer/data/gear.xml
```diff
@@ -1531,6 +1531,51 @@
...
+      <name>Ammo: Sling-Shot Paint Capsules</name>
+      <category>Ammunition</category>
+      <rating>0</rating>
+      <source>SOTG</source>
+      <page>13</page>
+      <avail>-</avail>
+      <cost>10</cost>
+      <costfor>10</costfor>
+      <weaponbonus>
+        <damage>0</damage>
```
DelnarErsike, line 1561: This will do nothing because `<damage>` just modifies the existing damage value of a weapon. So `<damage>0</damage>` means that the damage value is adjusted by 0, i.e. it remains unchanged.

Chummer/data/gear.xml
```diff
@@ -1531,6 +1531,51 @@
...
+      <name>Ammo: Sling-Shot Poison Capsules</name>
+      <category>Ammunition</category>
+      <rating>0</rating>
+      <source>SOTG</source>
+      <page>13</page>
+      <avail>-</avail>
+      <cost>100</cost>
+      <costfor>10</costfor>
+      <weaponbonus>
+        <damage>As Narcoject</damage>
```
DelnarErsike, line 1576: See my comment before about `<damage>` and `<damagereplace>` nodes.

DelnarErsike (changes requested):
Besides going in and changing the things I commented on that need changing, please make sure that the stuff you add actually works the way you want it to. With some of the stuff I commented, it would be immediately obvious that the current entry will not work the moment you just test it once yourself.

Chummer/data/weapons.xml
KoshiirRa: Fixed this one, this is what I get for copy/pasting from the wrong thing 🤣  Commit coming in a few.  No range is specified for this thing, I'm going to go out on a limb here and give it the same range category as the Ares Giantslayer Slingshot.

KoshiirRa:
Made requested changes.  Added bonus, I've realized that nobody has ever added ammo for the Ares Giantslayer, so now it has capsule rounds as described in its description too.

Chummer/data/gear.xml
```diff
@@ -1532,6 +1532,68 @@
...
+      <name>Ammo: Sling-Shot Acid Capsules</name>
+      <category>Ammunition</category>
+      <rating>0</rating>
+      <source>SOTG</source>
+      <page>13</page>
+      <avail>-</avail>
+      <cost>100</cost>
+      <costfor>10</costfor>
+      <weaponbonus>
+        <damagereplace>6P Acid</damagereplace>
```
DelnarErsike, line 1546: "Acid" should be in its own `<damagetype>` node. That way the 6P will properly be recognized as 6 damage of Physical type, allowing things like damage-modifying weapon accessories to work.

Chummer/data/gear.xml
```diff
@@ -1532,6 +1532,68 @@
       </weaponbonus>
       <ammoforweapontype>firefighting cannons</ammoforweapontype>
     </gear>
+	<gear>
+      <id>d71ea9c1-9b5f-428e-b849-7d665eb31c5e</id>
+      <name>Ammo: Sling-Shot Acid Capsules</name>
```
DelnarErsike, line 1537: Be consistent with the spelling of "slingshot"/"sling-shot". Either it has a hyphen everywhere or it is one word everywhere, don't mix up the two. This applies to all of the other slingshot ammo you added, though I will only make this comment for this entry.

Chummer/data/gear.xml
```diff
@@ -1532,6 +1532,68 @@
...
+      <category>Ammunition</category>
+      <rating>0</rating>
+      <source>SOTG</source>
+      <page>13</page>
+      <avail>-</avail>
```
DelnarErsike, line 1542: While Chummer can (now) handle an availability entry of `-`, it's better to use `0` so that its value can be processed as a number for things like the amend system. This applies to all of the other slingshot ammo you added, though I will only make this comment for this entry.

Chummer/data/gear.xml
```diff
@@ -1532,6 +1532,68 @@
...
+      <name>Ammo: Sling-Shot Paint Capsules</name>
+      <category>Ammunition</category>
+      <rating>0</rating>
+      <source>SOTG</source>
+      <page>13</page>
+      <avail>-</avail>
+      <cost>10</cost>
+      <costfor>10</costfor>
+      <weaponbonus>
+        <damagereplace>0</damagereplace>
```
DelnarErsike, line 1561: If you are doing a `damagereplace`, you should specify the type of damage, just to be on the safe side. That is why the T-Shirt Cannon has a damage code of `0S` instead of `0`.

Chummer/data/gear.xml
```diff
@@ -1532,6 +1532,68 @@
...
+      <name>Ammo: Slingshot Capsule Round</name>
+      <category>Ammunition</category>
+      <rating>0</rating>
+      <source>RG</source>
+      <page>23</page>
+      <avail>-</avail>
+      <cost>5</cost>
+      <costfor>10</costfor>
+      <weaponbonus>
+        <damagereplace>As Substance Within</damagereplace>
```
DelnarErsike, line 1591: "As Substance Within" is a term that won't get translated into other languages (note, neither will "Acid" for an earlier slingshot entry). "As Drug/Toxin", "Special", or "Chemical" would get translated though.

Chummer/data/weapons.xml
```diff
@@ -14843,6 +14843,30 @@
...
       <spec>Machine Guns</spec>
     </weapon>
+    <weapon>
+      <id>4629315e-9d09-4d04-92ac-3bad46675522</id>
+      <name>Ontario Arms Sling-Shot</name>
+      <category>Underbarrel Weapons</category>
+      <type>Ranged</type>
+      <conceal>6</conceal>
+      <accuracy>6</accuracy>
+      <damage>As Ammo</damage>
```
DelnarErsike, line 14853: This is a bit of a tricky entry here, especially because it won't get translated. If you absolutely do not want to set a "default" damage code, you should use "Special" here instead, as that will actually get translated.

Chummer/data/weapons.xml
```diff
@@ -14843,6 +14843,30 @@
...
+      <mode>SA</mode>
+      <ammo>10(m)</ammo>
```
DelnarErsike, line 14855: Are you sure this is correct, an internal magazine with a size of 10 for a slingshot?

DelnarErsike (commented):
Some stuff could use some tweaking, but it looks fine overall.

Chummer/data/weapons.xml
KoshiirRa: Good point, changed to "Special" instead.

Chummer/data/gear.xml
KoshiirRa: Changed to "As Drug/Toxin."

Chummer/data/gear.xml
KoshiirRa: Changed to 0 for all 4 of them.

Chummer/data/gear.xml
KoshiirRa: This one is a bit odd.  The capsule rounds are technically only listed for the Ares Giantslayer, which is printed as "slingshot" but the other 3 are for the Ontario Arms "Sling-Shot" with no mention of if they're usable with the Giantslayer.  Based on descriptions in both I'd say they are, so I'm going to change them all to "Slingshot."

Chummer/data/gear.xml
KoshiirRa: Fixed.

KoshiirRa:
Tweaks made.

DelnarErsike (approved):

PR #4720, opened 2022-03-02
add bonus dice to weapon and accessories

Stexinator:


ArchonMegalon:
I just took a look. What I don't get yet: BonusDice for ... what? Availability? Damage? 
Just calling it BonusDice and coding it for avail is not very generic.

Maybe a) call is BonusDiceAvail or b) only BonusDice but give it a property stating WHAT is enchanced.

chummer5a:
WeaponAccessory already has dicepool, there's no reason to add a second property for it. If you want to maintain the same name for both objects, suggest renaming WeaponAccessory.DicePool to SkillDicePool for both objects or similar.

Stexinator:
> WeaponAccessory already has dicepool, there's no reason to add a second property for it. If you want to maintain the same name for both objects, suggest renaming WeaponAccessory.DicePool to SkillDicePool for both objects or similar.

Great, that works as expected. The branch can be scratched.

Since this functionality is available smart gun systems can be reworked

PR #4809, opened 2022-05-27
Rework clip handling to fix bugs

TrueLunacy:
- Add and remove clips when clip-giving accessories are added
- Associate clips with clip-giving accessories, so they get removed when the accessory is removed
- Unload clips whenever they're being removed, also whenever the weapon is being removed
- Load accessories before loading clips, so we can access and associate them
- Fix bug when transferring ammo to vehicles, where the ammo wouldn't be transferred
- Remove AmmoSlots, it's not needed anymore

This fixes a number of issues reported in the Discord

There's still one outstanding bug I could find - the vehicle view doesn't update automatically when things change (ammo loaded, etc.). This doesn't affect functionality, just display

### DIFF
--- a/Chummer/Backend/Equipment/Clip.cs
+++ b/Chummer/Backend/Equipment/Clip.cs
@@ -28,16 +28,24 @@ namespace Chummer.Backend.Equipment
     {
         private readonly Character _objCharacter;
         private readonly Weapon _objWeapon;
+        // can be null for the internal clip
+        private readonly WeaponAccessory _objAccessory;
         private Gear _objAmmoGear;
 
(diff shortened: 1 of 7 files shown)


TrueLunacy:
Second commit adds a checkbox to filter out loaded ammo.

I don't know the procedure for things like other languages, if you need to add a default string or something. Nor do I know how changelog.txt is meant to be updated - do you update it when you make the changes, or when a release will happen, or so on.

This adds one new string - Checkbox_HideAmmo - for the new filter checkbox.

ArchonMegalon:
Hi @TrueLunacy !

Can you please take a look at this?

https://ci.appveyor.com/project/chummer5a/chummer/build/5.222.118-002248-failed

I guess your commit is more likely to have caused this than mine...

PR #4825, opened 2022-06-11
Fix a few ammunition bugs.

TrueLunacy:
First commit: Fixes an issue commented in Discord, where ammunition can be removed/added/merged/split while loaded into a clip.
Ammunition can still be sold when loaded into a clip, but this just unloads the clip.

Second commit: Fixes reloading taking too much ammo out of some ammunition, leading to negative ammo values

### DIFF
--- a/Chummer/Backend/Equipment/Weapon.cs
+++ b/Chummer/Backend/Equipment/Weapon.cs
@@ -6541,7 +6541,7 @@ namespace Chummer.Backend.Equipment
                         else
                         {
                             objCurrentlyLoadedAmmo.Quantity = decQty;
-                            objSelectedAmmo.Quantity -= decQty;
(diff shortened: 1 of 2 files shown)


chummer5a:
All looking good so far, aside from the PSG Enforcer and Beretta Northstar generating an error dialog when selected in debug mode. I'm assuming this is just due to the debug.assert and that can be removed?

TrueLunacy:
The debug.assert can be removed, but it exposes a bug - I put it in there to ensure the precondition (_lstAmmo was empty for CreateClips()), which it caught!

Just removing it leads to the PSG Enforcer having 3 clips - Create() added the clip accessory, which triggered the observable to add a clip, then CreateClips() goes to create three clips. But the fix for this is fortunately easy, just clear _lstAmmo before calling CreateClips().

Issue #4857, opened 2022-07-12
Updated Shamanic Traditions are Missing


darksnakezero:
**Is your feature request related to a problem? Please describe.**
A) When selecting a tradition of a mage, I can not select the updated traditions Shamanic [Traditionalist] and Shamanic [Ancestor]. They dont show up in the List.
B) If I select Shamanic [Traditionalist], I want to automatically gain the spirit affinity PQ when reaching the required Initiate Level 3.
C) If I select Shamanic [Traditionalist], the prerequisite should be Mentor spirit.

**Describe the solution you'd like**
The two traditions show up in the list. 

**Describe alternatives you've considered**
As Chummer does not currently support possibilities to Enforce B and C, I still would like to be able to select the traditions and have the Binding skill deactivated in Case Shamanic [Traditionalist].

I ll create a merge request with the apropriate solution to solve Part A of the problem.

**Additional context**
The Traditions are described on FA on pg. 73-74.

PR #4858, opened 2022-07-12
Added the traditions Shamanic[Traditional] and Shamanic[Ancestor]. I …

darksnakezero:
Partly Fixes: Updated Shamanic Traditions are Missing #4857 

[x] Added the traditions Shamanic[Traditional] and Shamanic[Ancestor]. 

I dont see how I could implement the requirement for Shamanic[Traditional] to have the mentor spirit. It clearly requires the positive Quality but does not give it for free. 
-> Someone creating a Shamanic[Traditional] character will have to satisfy the requirement of mentor spirit namually. 

Also, there is no way to automatically give a positive Quality on reaching a certain initiate level. 
-> This also has to be done manually.

### DIFF
--- a/Chummer/data/traditions.xml
+++ b/Chummer/data/traditions.xml
@@ -941,6 +941,39 @@
         <spiritmanipulation>Spirit of Earth</spiritmanipulation>
       </spirits>
     </tradition>
+    <tradition>
+      <id>9d18592e-5f49-4992-babd-d1ac98c48f78</id>
+      <name>Shamanic [Traditionalist]</name>
+      <drain>{WIL} + {CHA}</drain>
+      <source>FA</source>
+      <page>74</page>
+      <addqualities>
+        <addquality forced="False">Mentor Spirit</addquality>
+        <addquality forced="True" select="Shamans Code: Harmony with Nature">Code of Honor</addquality>
+      </addqualities>
+      <spirits>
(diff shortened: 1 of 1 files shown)


Chummer/data/traditions.xml
```diff
@@ -941,6 +941,39 @@
...
       </spirits>
     </tradition>
+    <tradition>
+      <id>9d18592e-5f49-4992-babd-d1ac98c48f78</id>
+      <name>Shamanic [Traditionalist]</name>
+      <drain>{WIL} + {CHA}</drain>
+      <source>FA</source>
+      <page>74</page>
+      <addqualities>
+        <addquality forced="False">Mentor Spirit</addquality>
```
chummer5a, line 951: They don't GET the quality, do they? I read that more as a requirement than a benefit.

Teksura: > They don't GET the quality, do they? I read that more as a requirement than a benefit.

The exact text in the book says:

> Must take the Mentor Spirit quality.

I read it the same as you. they do not seem to just get Mentor Spirit for free. Mentor Spirit looks more like it is a prerequisite.

Teksura:
We could prolly push this if we switch out the addqualities for a required. Do you must have the Mentor spirit Quality in order to qualify for the tradition.

PR #4943, opened 2022-10-17
Overhaul pre-commit hook

davidfoerster:
* More robust file status filtering using 'git ls-files'
 * which avoids the non-portable use of 'findstr'.
 * Use a pipe plus while-loop over command subsitution plus for-loop.
 * Get rid of useless recursive invocation of 'grep'.
 * Use printf how it's supposed to be used to make it more robust: with argument format specifiers.
 * Make the script file executable.

Among other benefits, this should enable *nix users to submit pull requests for translation and data changes.

### DIFF
--- a/git/hooks/pre-commit
+++ b/git/hooks/pre-commit
@@ -9,16 +9,11 @@
 #set -x
 
 exitstatus=0
-git ls-files --modified --exclude-standard ':/**/app.config' |
+git grep --cached --files-with-match -e '<nlog[\s>]' -- ':/**/app.config' |
 while read -r f
 do
-	if grep -q "<nlog>" "$f"
-	then
-		printf 'Please remove the nlog-block in file %s!\n' "$f"
-		exitstatus=1
-	else
-		printf 'NLog not found in %s.\n' "$f"
-	fi
+	printf 'Please remove the nlog-block in file %s!\n' "$f"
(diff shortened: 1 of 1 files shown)


git/hooks/pre-commit
```diff
@@ -8,25 +8,17 @@
 # To enable this hook, rename this file to "pre-commit".
 #set -x
 
-List=$(git status | findstr modified | findstr app.config | cut -d ":" -f2 | xargs)
 exitstatus=0
-
-for f in "$List"
+git ls-files --modified --exclude-standard ':/**/app.config' |
```
DelnarErsike, line 12: I wasn't aware of the `git ls-files` command before, and it seems very handy for this case.

However, it seems like `git ls-files --modified` won't catch files that are newly added to the repository. This could create a niche issue where someone creates a new project in the solution, has the `app.config` of the project contain an undesired block, and then can commit it to the repo without the pre-hook catching it.

davidfoerster: I hadn’t thought of that. Good catch!

The simplest solution I can think of is to `git grep` all cached files for forbidden patterns.

davidfoerster:
Wait. I’ll need to make one more change. Since the loop is connected to a pip it runs in a subshell and the parent process won't see the change to the exit status.

PR #4950, opened 2022-10-26
NeonAnarchy House Rules

Djordje-Nikolic:
We've updated the Neon Anarchy customdata and settings in preparation for the server reset in a few days time.


Chummer/customdata/Neon Anarchy House Rules/amend_qualities.xml
```diff
@@ -230,14 +250,23 @@
...
+                    <quality>Albinism II</quality>
+                    <quality>Allergy (Common, Moderate)</quality>
+                    <quality>Astral Beacon</quality>
+                    <quality>Blank Slate</quality>
+                    <quality>Incomplete Deprogramming</quality>
+                    <quality>Insomnia (Basic)</quality>
+                    <quality>Low Pain Tolerance</quality>
+                    <quality>Sensitive System</quality>
+                </selectquality>
+            </bonus>
```
DelnarErsike, line 268: Just to be clear, `amendoperation="replace"` completely replaces the entire existing node with the entire new node. Having another node with `amendoperation="replace"` inside of it makes no sense, and probably signals that you want to do something that may not work the way you actually want it to work.

Chummer/customdata/Neon Anarchy House Rules/amend_qualities.xml
```diff
@@ -275,43 +304,19 @@
...
 		<quality>
 			<name>Trust Fund IV</name>
-			<required amendoperation="replace">
-				<oneof>
-					<quality>SINner (National)</quality>
-					<quality>SINner (Corporate)</quality>
-					<quality>SINner (Corporate Limited)</quality>
-				</oneof>
-			</required>
+			<hide />>
```
DelnarErsike, line 319: Double closing `>>` almost certainly will cause problems. You try to find and remove any and all of these (I don't have time to do a full sweep myself right now). I'd recommend you use a text editor with automatic tag completion like VSCode if you struggle with this in the future.

DelnarErsike (changes requested):
There are a few small issues in some of the changes you've made that you'll want to address before merging.

Chummer/customdata/Neon Anarchy House Rules/amend_qualities.xml
Djordje-Nikolic: I believe this was solved in the latest commit.

Chummer/customdata/Neon Anarchy House Rules/amend_qualities.xml
Djordje-Nikolic: I believe this was solved in the latest commit.

PR #4963, opened 2022-11-26
Update gear.xml

Teksura:
Adds Street Cooked, Pharmaceutical, and Designer drug grades as plugins to Drugs.

Adds <addoncategory>Drug Grades</addoncategory> to all Drugs to restrict the plugin menu to only show the Drug Grades as a plugin so that menu is reasonably user friendly.

Also sets Psychchips (Legal) and Psychchips (Illegal) to be categorized as BTLs instead of Drugs. No idea why they were set to Drugs.

~~Todo in the future: Figure out a way to make Pharma/Designer/Street Cooked ONLY show up in Drugs, so people can't make, like, a Street Cooked Ambulance Gurney. I'll get right on that when I can be bothered to get right on that. So, possibly some time next year.~~

### DIFF
--- a/Chummer/data/gear.xml
+++ b/Chummer/data/gear.xml
@@ -49,7 +49,7 @@
     <category blackmarket="Electronics">Cyberdecks</category>
     <category blackmarket="Electronics">Cyberterminals</category>
     <category blackmarket="Electronics">Disguises</category>
-    <category blackmarket="Drugs">Drugs</category>
+    <category blackmarket="Drugs"> </category>
     <category blackmarket="Electronics">Electronics Accessories</category>
     <category blackmarket="Electronics">Electronic Modification</category>
     <category blackmarket="Electronics">Electronic Parts</category>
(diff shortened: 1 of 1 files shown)


Teksura:
Added <required> to the drug plugins so now you can't make the Street Cooked Ambulance Gurney anymore.


Joschi did point out there is an edge case where someone may have enough money to buy a street cooked drug but not enough for the standard drug and technically *cant* buy it because they'd have to buy the standard before they make it street cooked. I have no good answer for that.

Chummer/data/gear.xml
```diff
@@ -13769,6 +13839,66 @@
...
+      <id>f22e79fa-fb04-4369-a761-a1a46e242bc8</id>
+      <name>Street Cooked</name>
+      <category>Drug Grades</category>
+      <rating>0</rating>
+      <source>CF</source>
+      <page>190</page>
+      <avail>0</avail>
+      <capacity>0</capacity>
+      <requireparent />
+      <cost>Parent Cost * -.5</cost>
```
DelnarErsike, line 13855: This is a bit dicey because it can create situations where someone can afford to buy a Street Cooked grade drug but not the Standard grade one. This is because the process of buying gear with some plugin after character creation is that you buy the base gear first, get the cost of the base gear deducted from your nuyen, and then buy the plugin and have its cost deducted from your nuyen. A bit of an edge case scenario, but still worth noting.

PR #4965, opened 2022-12-05
Groundwork for the 4e Critter Custom Data File

Teksura:
Laying out the groundwork for the 4e Critter Custom Data File. The First Step is to remove ALL of the 4e critters from the ungodly mess that is critters.xml and while I'm at it sort the damned thing so it is less of an ungodly mess. It's still an ungodly mess, but slightly less so now.

All 4e critters are now moved to a totally non-functional amends file. This will serve as the groundwork for a larger project to slowly start converting the critters to be functional in Chummer5 if someone wishes to enable the amends and use the critters based on their 4e stats.


Teksura:
Since there is A LOT of stuff going on, I want to clarify I checked what I was doing to ensure I fixed any cock-ups before pushing.

While I did not go opening every single critter to confirm it was still working as intended, I did open two instances of chummer to verify that every critter which was available in the previous version of critters.xml was still showing up in this version of critters.xml

Screen Shot 2022-12-04 at 6 25 18 PM

This update helps make critters.xml a lot less of an ungodly mess by actually sorting the blasted thing, shifts the unused 4e critters into an amends where they won't bother anyone, and I am fairly certain it manages to not break anything in the process.

I hope.

PR #4991, opened 2023-02-07
Giving Quick Healer a +2 bonus to First Aid and Medicine

Teksura:
Quick Healer's text reads:

> A character with the Quick Healer quality receives a +2 dice pool modifier to all Healing Tests made on/for/by her, including magical healing.

As it provides a bonus to tests made by the character, this would apply to First Aid and Medicine tests. I am unaware of any RAW tests involving the First Aid or Medicine skill which are not Healing tests.

### DIFF
--- a/Chummer/data/qualities.xml
+++ b/Chummer/data/qualities.xml
@@ -521,6 +521,14 @@
           <val>2</val>
         </spelldicepool>
         <stuncmrecovery>2</stuncmrecovery>
+        <specificskill>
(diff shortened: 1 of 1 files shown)


Stexinator:
eg Diagnosis-Tests

DelnarErsike:
Ugh, the advanced First Aid rules from Bullets & Bandages, yes. You will probably need to take those into account.

Teksura:
Arrrug. You're right.

PR #4996, opened 2023-02-24
Update French translations and page numbers for armors

gdelmee:
Correct translations and pages using official books.
For non officially translated books, I've kept the current page of the English version?


DelnarErsike:
It's missing the no-width spaces in front of apostrophes (this is to fix issues that WinForms has with alphabetical ordering where it doesn't know how to properly deal with French apostrophes), but I can handle those.

PR #4998, opened 2023-02-25
Add some tweaks to the printed character sheet

MachineMuse:
* Divide qualities into Positive and Negative sections
* Group Metagenic qualities together and label them accordingly (since this information isn't exposed, it instead checks whether the quality is from RF pages 110-124)
* Users can add a a `Prefix:` line at the end of each quality's Notes to add a prefix which will additionally sort by (e.g. so Magical qualities can be grouped separately from skill-related qualities)
* Include gear notes
* For gear column balancing purposes, include notes and children in line count - it's impossible to perfectly balance them since we have no way of detecting when a line wraps, but we can use this as a slightly better approximation

### DIFF
--- a/Chummer/sheets/Shadowrun 5 set.xslt
+++ b/Chummer/sheets/Shadowrun 5 set.xslt
@@ -622,17 +622,31 @@
                 <td style="text-align: center;">
                   <xsl:if test="qualities/quality">
                     <div class="block" id="QualitiesBlock">
-                    <table class="tablestyle">
-                      <tr>
-                        <th width="80%" style="text-align: left">
-                          <xsl:value-of select="$lang.Quality" />
-                        </th>
-                        <th width="10%" />
-                        <th width="10%" />
-                      </tr>
-                      <xsl:call-template name="Qualities" />
-                    </table>
-                  </div>
+                      <table class="tablestyle">
+                        <tr>
+                          <th width="80%" style="text-align: left">
+                            <xsl:value-of select="$lang.Qualities" />: 
+                            <xsl:value-of select="$lang.Positive" />
+                          </th>
+                          <th width="10%" />
+                          <th width="10%" />
+                        </tr>
+                        <xsl:call-template name="Qualities">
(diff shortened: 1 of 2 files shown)


MachineMuse:
Example output:
image

image

DelnarErsike:
I am not really satisfied with the implementation of the changes here. My issues mainly stem from a code maintenance perspective.

- Determining if a quality is metagenic by filtering for those with Run Faster as a sourcebook and a specific range of pages is very poor form. Anyone who modifies any of those listed qualities to not be metagenic via amends, who adds their own metagenic qualities that are not in Run Faster, or who might potentially print in a language where page numbers for qualities are potentially shifted up or down will not be covered. Besides, just exposing the actual property to printing/export is trivial, see commit 089e92a0bd7451b19250a0eee3fafe8db3ee95c3, after which you can just filter based on that exposed property.
- Your label for "(metagenic)" and the conceit of the system based on using "Prefix" in notes fields is defined in and revolves around English. Chummer not only supports multiple languages, but supports printing in languages different from the UI language. Even if a keywording system were a good idea (I'll get to that later), you'll want to support every version of that keyword that would seem straightforward from a user perspective: uppercase, lowercase, title case, space before the colon (as is custom in French), mono-width versions of the previous (as is custom when typing in Latin characters in an East Asian language), etc.
- Relying on keywords in something's notes value by which to sort items in one of the character sheets is not just very hacky, but is horrible from a UX perspective, especially for all users who are not actively aware of this feature. Notes fields are intended to be just that, notes. They are not intended to have any functionality across any items whatsoever. This is an unsaid expectation, but it is a vital one to uphold. The fact that users can toggle the exporting of their notes to their sheet in the first place is an extension of this expectation (and a keywording system that produces a sort order based on contents in notes fields would also get shut down by users simply toggling that off). If you want to add the ability for users to auto-sort items on their sheet in various ways, there are at least a handful of ways to do that which do not involve hijacking exposed parameters that are not intended to be used to determine sort order.

PR #5051, opened 2023-09-06
Implemented an ApplicationMonitor Class to kill Chummer from a timer thread if needed.

### LINKED ISSUE #5010: Chummer not actually exiting when closed
5.225.144 When hitting the x to close out the app, chummer is remaining open in the task manager. The window closes and it appears closed but trying to start chummer again fails unless I kill the remaining chummer task in task manager. Noticed on windows 10.

JoschiZ:
fixes #5010

Implemented an `ApplicationMonitor` class that starts a timer on another Thread.
If the set interval elapses `Application.Exit()` will be called to stop the WinForms event loop.
In the case of Chummer still not closing properly and progressing beyond `Application.Run()` we kill the process on the next timer tick.

The timer is started as the last step of the Main Form closing.
The timer is stopped if we progress beyond Application.Run() which implies that we now started to gracefully exiting Chummer.

### DIFF
--- a/Chummer/Backend/Helpers/ApplicationMonitor.cs
+++ b/Chummer/Backend/Helpers/ApplicationMonitor.cs
@@ -1,0 +1,59 @@
+using System;
+using System.Windows.Forms;
+using NLog;
+
+namespace Chummer
+{
+    /// <summary>
+    /// Class to handle the forceful shutdown of Chummer if the eventloop stucks for some reason.
+    /// </summary>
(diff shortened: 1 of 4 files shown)


DelnarErsike:
This is a bit of a hacky fix in general. Do we still not know why Chummer isn't closing properly in the first place?

DelnarErsike:
This PR is no longer necessary. Closing it.

PR #5053, opened 2023-09-10
Setup Chummer as .NET7 WinForms Blazor Hybrid

JoschiZ:
This pull request moves Chummer and all it's dependencies to .NET 7 and sets it up to be used as an WinForms Blazor Hybrid application. As an example the About page got re implemented in Blazor following the MVVM principle, with all Blazor related files being isolated in a `Razor Class Library` to enforce a separation of the `Views` and `ViewModels` as much as possible.

The purpose of this is to allow a gradual move away from WinForms and towards newer platforms like MAUI.

Besides changes that are needed for a .NET7 conversion the following notable decisions / changes are included:
- [MudBlazor](https://mudblazor.com/) was used as an free MIT licensed UI Library, which is based on Material Design.
- Unit tests got rewritten to fully use xUnit instead of MSTest, mainly because I could not get them to run in .NET7.
- Blazor was set up with MVVM in mind using the [MVVM Tookit](https://learn.microsoft.com/en-us/dotnet/communitytoolkit/mvvm/)
- The CodeQuality Analyzers are now included by default and got configured to use the newest [recommended](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/overview?tabs=net-7#enable-additional-rules) rules set on all projects.

There are a few tasks left, before this is should be merged. See the [TODO File](dotnet7_update_todo.md) for a comprehensive list of tasks, with comments. Below are just the most important ones.

# .NET7 Update TODO

## Need to be done before merging
- [ ] Ensure that the automated build process still works (I will need some help/direction with this)
  - After digging through the appveyor.yml it should still work, because it's framework agnostic. But I can't actually test it without setting up my own appveyor.
- [x] Exclude ``Chummerhub.Client`` from the build
- [x] Check and update the README

## Should be done before merging
- [x] Test Linux support
  - [ ] Update README
  - NOTE: Sadly this will cost us Linux (tested on Ubuntu with Wine) support for now, because WebView2 is not working right now. There may be ways to get it running anyways. Linux support is on the roadmap for WebView2, but it's probably not coming any time soon.

- [ ] Create a new ``.NET Framework`` branch and archive it right before the merge.

## Can / should be done later
-  Address the new ``CodeQuality.Analyzer`` warnings.
-  Replace obsolete APIs
  -  `WebClient` with `HttpClient`
  -  `new TelemetryClient()` with `new TelemetryClient(TelemetryConfiguration config)`.
  -  `Logmanager.GetCurrentClassLogger()`
-  Implement translations for `System.Private.CoreLib` exceptions.

# Additional improvements
Here I list improvements that are made possible in .NET7, that we should take advantage on after the merge.
Some may already be done, but I leave them out of this PR for now to keep it as focused on the framework conversion as possible. (Some minor things may have snug in)

- [x] Source Generated Regex (https://github.com/JoschiZ/chummer5a/tree/dotnet7-move-to-regex-sourcegen)
- Replace DllImport with LibraryImport taking advantage of the [P/Invoke](https://learn.microsoft.com/en-us/dotnet/standard/native-interop/pinvoke-source-generation) Source generator. This is not a complete drop-in replacement for everything. Because we don't actually seem to be using native methods that much I don't see this as being a high prio task.


ArchonMegalon:
Wow - that reads awesome. 
Don't get yourself bogged down because of the ChummerHub-Plugin (or even the Pluginframework as a whole). Just remove it for now and move forward!

JoschiZ:
Thanks,

I updated the Task list of the PR, feel free to expand on it.

One problem I face is, that I don't actually know how releases and nightlies are done, because of that I can't really test them.
But besides that it's just some housecleaning and busywork. 

But the main TODO now, would be to get Chummer ready for the next release.

JoschiZ:
I will close this for now and pick it up again at a later stage when we actually want to proceede with this move

PR #5055, opened 2023-09-21
Cybersuite children are now included in ATT pools

### LINKED ISSUE #5045: Attribute calculations do not take cyberlimbs in cybersuites into account
Chummer does not properly calculate a character's physical attributes if they have gotten cyberlimbs from a cybersuite.

To reproduce: create a character in chargen, then give them the Evo Better than Human. Although the character properly gains the +4 physical condition monitor boxes, the cyberlimbs' average STR/AGI is not used to calculate the character's STR/AGI, and the dice pools for, e.g., Unarmed Attack do not use the cyberlimb STR/AGI either.

The expected behavior is that cyberlimbs in a cybersuite will function identically to cyberlimbs outside of a cybersuite for all calculations.

JoschiZ:
This is done by allowing the cybersuite to inherit attributes of their children and raising a ATT refresh when a cybersuite is added or removed. The relevant tooltips are updated to recursively add the children of Cybersuites.

fixes #5045

### DIFF
--- a/Chummer/Backend/Attributes/Attribute.Core.cs
+++ b/Chummer/Backend/Attributes/Attribute.Core.cs
@@ -1023,7 +1023,7 @@ namespace Chummer.Backend.Attributes
                     Cyberware.CyberlimbAttributeAbbrevs.Contains(Abbrev))
                 {
                     return _objCharacter.Cyberware.Any(objCyberware =>
-                        (objCyberware.Category == "Cyberlimb" || objCyberware.Category == "Cybersuite") && !string.IsNullOrEmpty(objCyberware.LimbSlot), token);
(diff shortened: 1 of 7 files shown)


DelnarErsike:
I would honestly prefer a solution that is more general than this one. The main thing I'm worried about is that the fix specifically requires that (and relies on) the parent 'ware be of the "Cybersuite" category. This shouldn't necessarily have to be the case. Hardcoding around the "Cyberlimb" category already causes a few problems, we don't want to make this worse.

JoschiZ:
I extracted all but one check for "Cyberlimb" and replaced them by a single centralized property on the `Cyberware` itself, that checks if itself or any of it's children occupies any limb slots.
This should be functionally equivalent and will allow easier refactor in the future.

PR #5081, opened 2023-11-14
Specifying the Multiple Opponent Defense of La Verdadera Destreza

IsaacAlpharn:
Specifying the Multiple Opponent Defense of La Verdadera Destreza in German printings of Run and Gun.

image

image


DelnarErsike:
Do you want to modify the solution files so that Chummer includes the new file with its builds, or do you want me to do it after merging the PR?

IsaacAlpharn:
After merging, I think.

PR #5084, opened 2023-11-20
Attune Item Support

### LINKED ISSUE #5078: [Feature Request] An "Attuned Weapon" Weapon Modification to keep track of attuned weapons.
Although it is possible to manually keep track of attuned equipment for Adepts/Mystic Adepts, it would be preferable in my opinion to have an "Attuned Item" weapon modification available for all weapons to do so automatically. I am not sure if it would be possible for these attachments to reference the character's initiate grade, but if not, it would be possible to implement as a weapon modification that takes up no slots, is available for all weapons, and has an arbitrarily high maximum rating, adding +1 Accuracy per rating.

If it is possible to reference the character's initiate grade, then it could simply reference that and add that number to Accuracy.

Teksura:
Adding support for Attune Item in a way that makes me furious but technically works. 

If anyone does somehow end up in a situation where they need this at IG 13 or above, congratulations... Okay, fine. Just take the 12+ accessory and then as many +1 accessories as you need. Don't @ me though because this is more your fault than mine WTF are you even doing at IG 13?

Resolves #5078 I guess. But I'm still not happy with the way this works.

### DIFF
--- a/Chummer/data/weapons.xml
+++ b/Chummer/data/weapons.xml
@@ -19066,20 +19066,267 @@
     </accessory>
     <!-- End Region -->
     <!-- Region Attune Item -->
-    <accessory>
+     <accessory>
       <id>0fcf55fa-afb4-487c-93f2-f41d6fc4e769</id>
-      <name>Attuned Item</name>
+      <name>Attuned Item: (IG 1)</name>
       <mount />
       <avail>0</avail>
       <cost>0</cost>
       <source>SG</source>
       <page>124</page>
-      <rating>0</rating>
       <accuracy>1</accuracy>
       <required>
-        <oneof>
+        <allof>
           <spell>Attune Item</spell>
+          <initiategrade>1</initiategrade>
+        </allof>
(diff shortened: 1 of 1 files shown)


DelnarErsike:
I'm not super-happy with this solution because it's *extremely* hacky. I'd honestly much prefer if this was something implemented in the C# side, e.g. by incorporating initiate and/or submersion grade as a dedicated keyword that could be fetched for calculations the same way attributes can.

Teksura:
> I'm not super-happy with this solution because it's _extremely_ hacky. I'd honestly much prefer if this was something implemented in the C# side, e.g. by incorporating initiate and/or submersion grade as a dedicated keyword that could be fetched for calculations the same way attributes can.

I agree. My skills however are limited to XML fuckery. And good *god* this is some fuckery. So I did the best my abilities could muster.

Your described solution would be far superior and would be able to apply in other situations as well. Perhaps something like being able to use `{IG}` or `{SG}`.

PR #5086, opened 2023-12-02
DNArt & Chimeric Pets

Teksura:
Adds DNArt & Chimeric Pets from Chrome Flesh p 167

image

Costs are set to Variable because that + isn't clear beyond setting a minimum, so, I set that as the minimum with a maximum of a million similar to Custom Item. 

If you're playing a game where you have enough spare funds around to blow more than a million NuYen on a bit of DNArt, congratulations.

### DIFF
--- a/Chummer/data/gear.xml
+++ b/Chummer/data/gear.xml
@@ -21026,6 +21026,47 @@
       <cost>20</cost>
     </gear>
     <!-- End Region -->
+    <!-- Region Chrome Flesh Transgenetic Art and Chimera Pets -->
+    <gear>
+      <id>27b41217-6580-4ed8-8033-09b3b31075b7</id>
+      <name>DNArt (Commercial)</name>
+      <category>Custom</category>
+      <rating>0</rating>
+      <source>CF</source>
+      <page>167</page>
+      <avail>10</avail>
+      <cost>Variable(30000-1000000)</cost>
+    </gear>
+    <gear>
+      <id>1d3ab826-04dc-4d5c-acf6-2c6f39580118</id>
+      <name>DNArt (Custom)</name>
+      <category>Custom</category>
+      <rating>0</rating>
+      <source>CF</source>
+      <page>167</page>
+      <avail>20</avail>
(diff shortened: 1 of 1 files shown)


Teksura:
> These are transgenics, so I assume they would need to work as genemods, correct? Why are they gear items?

Genemods would be applying the genetic modification to the character as they are added via the Cyberware & Bioware tab. But this is not that, and not a modification to the character. This is basically buying a plant or a cat which has been genetically modified and now you own an item which is a plant or a cat or something which, to quote Chrome Flesh p 167:

> The fabulously wealthy have always used exquisite- ly bred animals as status symbols and in many cases as living fashion accessories. Genecrafted plants and an- imals bring this to a whole new level. The latest trend in fashion pets are PetitesTM, tiny chimeric animals cre- ated to look like larger natural animals—the larger, the better. The most common Petites are giraffes, elephants, and polar bears, none of them bigger than a lap dog. Another trend is BiowearTM, which are living articles of clothing. They often are sensitive to their surroundings, changing color, texture, or bioluminescing depending on the pheromones and sounds in their environment. Among the wealthiest of the wealthy one can find ar- boretums full of chimeric plants and animals (or some things in between) that outright defy the process of nat- ural evolution.

DelnarErsike:
Hmm, this is tricky. I'd categorize these mods then as similar to critter cyberware that's statted out in Howling Shadows, but IIRC, those are handled by Chummer as cyberware. Yet, obviously, it's not the critter pet that's going to be paying for those or these, it's the pet's owner.

Teksura:
Some of them are also plants, which chummer doesn't even really support. And they don't even really appear to have any associated essence cost with them either so it would be strange and awkward to try and make them critter cyberware because if a character wants to buy a fern that glows in the dark and changes colors, first they have to make an amends file to make a Fern critter, then they have to modify the Fern with the Cyber/Bioware tab. And they still have to pay for it on their Character's file.

I consider this to be in the same situation as the Savior Medkit and similar nanogear. Yes, it's nanoware. But it's not added to the character under the Cyberware options like you'd add a nanosystem.

PR #5163, opened 2024-10-29
Dealer Connection Language Fix

Teamozuka:
Dealer Connection wouldn't work for German and probably every other language thats not English because "Drones" != "Drohnen"


ArchonMegalon (approved):

chummer5a:
I'd prefer to see the improvement changed to apply the english string as selectedvalue rather than the current language and have the method in vehicle check for both; dunno why the improvement got set like that in the first place though.

DelnarErsike:
This is a bad fix for the issue. The issue specifically is that Dealer Connection saves the category to which it applies in the user's local language, which means that recognition of it fails later on. This commit changes recognition to work off of the user's local language as well, but this ends up breaking saves where Dealer Connection was added to a character in a UI language different from the current UI language. An actual fix would be to change the Dealer Connection improvement itself to save in English (or some universal key) so that the resulting character data (and save file) are agnostic of the current UI language.

PR #5164, opened 2024-10-29
Custom Drug Creator Fixes

Teamozuka:
Fixed a bunch of issues in the custom drug creator.
1. Got rid of a double colon after speed.
2. Made it correctly calculate grade cost.
3. Fixed Duration (used to always show "immideate")
4. Fixed a bug where it would show wrong stats after the reload calculation because it didnt save the level of componenteffects meaning all effects of a component were level 1(index 0) for the program. Adding a level 1 block then made him add the stats of every level while adding any other lever did nothing at all.

Before
custom drug

After
custom drug fix

### DIFF
--- a/Chummer/Backend/Equipment/Drugs.cs
+++ b/Chummer/Backend/Equipment/Drugs.cs
@@ -498,13 +498,13 @@ namespace Chummer.Backend.Equipment
         /// <summary>
         /// Total cost of the Drug.
         /// </summary>
-        public decimal TotalCost => Cost * Quantity;
+        public decimal TotalCost => Cost * Quantity * Grade.Cost;
 
         /// <summary>
         /// Total cost of the Drug.
         /// </summary>
         public async Task<decimal> GetTotalCostAsync(CancellationToken token = default) =>
-            await GetCostAsync(token).ConfigureAwait(false) * Quantity;
+            await GetCostAsync(token).ConfigureAwait(false) * Quantity * Grade.Cost;
 
         public decimal StolenTotalCost => Stolen ? TotalCost : 0;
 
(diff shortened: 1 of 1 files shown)


Chummer/Backend/Equipment/Drugs.cs
```diff
@@ -1264,13 +1264,16 @@ public async Task<string> GenerateDescriptionAsync(int intLevel = -1, bool blnEf
...
 
                     if (Category == "Custom Drug" || Duration != 0)
-                        sbdDescription.Append(await LanguageManager.GetStringAsync("Label_Duration", strLanguage, token: token).ConfigureAwait(false))
-                                      .AppendLine(DisplayDuration);
+                        sbdDescription.Append(await LanguageManager.GetStringAsync("Label_Duration", token: token).ConfigureAwait(false))
+                                      .Append(strSpace)
+                                      .Append("10 ⨯ ")
+                                      .Append((Duration + 1).ToString(GlobalSettings.CultureInfo))
+                                      .Append(await LanguageManager.GetStringAsync("String_D6", token: token).ConfigureAwait(false)).Append(strSpace)
+                                      .AppendLine(await LanguageManager.GetStringAsync("String_Minutes", token: token).ConfigureAwait(false));
```
DelnarErsike, line 1272: I would prefer it if the DisplayDuration property itself (and/or a newly-added async getter version to support `LanguageManager.GetStringAsync`) were fixed instead of only applying the fix to the specific part where it is used to generate the drug's description. There are probably other places where getting the displayed version of a drug's duration might be useful beyond just the description block in the UI that's meant to show all information.

Chummer/Backend/Equipment/Drugs.cs
```diff
@@ -790,8 +790,8 @@ public int Duration
             {
                 if (_intCachedDuration != int.MinValue)
                     return _intCachedDuration;
-                if (string.IsNullOrWhiteSpace(_strDuration))
-                    return _intCachedDuration = 0;
+                //if (string.IsNullOrWhiteSpace(_strDuration))
+                //    return _intCachedDuration = 0;
```
DelnarErsike, line 794: Rather than disabling this shortcut version of caching drug duration when there are no extra duration-modifying blocks applied, I think it would be better to just make sure it properly detects when there is indeed nothing present that would require expensive duration calculations to be done. Assuming you can do that fast enough, of course; if you can't then yeah, just remove this code.

PR #5166, opened 2024-11-08
Fixes GUID on the 4e Critter manifest

Teksura:
Custom Data manifests for "4e Critters Conversion" and "Adapsin applied as mult, not grade" have the same GUID, which causes Chummer to ignore the Adapsin CD


DelnarErsike:
Made irrelevant by commit b62e9dd905382604ac3ac585a08b350ad196e239.

PR #5171, opened 2024-12-23
Made it so Settings folder appears prior to first launch

Teksura:
I have found many new players in my LC get tripped up when downloading chummer and installing amends and settings because they skip the step of "launch chummer to generate settings folder" and get confused when they can't find the settings folder.

GitHub does not allow the creation of empty folders. This was resolved with the tried and true `coconut.jpg`

Suffer.


Teksura:
At the suggestion of truelunacy on the Discord, a more sane option was pursued. 


The goal here is simply to put *something* in a settings folder so the initial instal of Chummer has the Settings folder already there, thus eliminating the need to launch and close chummer after a fresh instal to generate the settings folder. (Or the alternative of create the settings folder manually).

PR #5182, opened 2025-04-05
Fix misspelling in critter.xml

BM123499:
Hi guys. It's my first time here, I found some misspelling in this file. I don't know if `)` is meant in some context, but I removed either way. If I need extra steps, please tell me.


Teksura:
Found in line 25897:

```<complexform>Deezz</complexform>```

We're all thinking it now.




Seriously though, are you sure about your removal of the karma cost of metavariants?  I thought that was correct. Can someone verify that?

BM123499:
I not removing the `karma` value of each metavariant, it already has a karma value, which is the karma value when you build the character with karma points instead of priority. I deleted the karma cost of priority builds, that is already incremented in `priority.xml`

BM123499:
see #5187 5187

PR #5187, opened 2025-04-23
[WIP] Recreate all XSD files

BM123499:
It's been over three years since the XSD files were last updated. While they're not strictly required for the application to function, keeping them updated can help with validation and provide a more standardized structure for defining variables.

I'm currently working on an importer that brings this dataset into Foundry VTT. In doing so, I've encountered a number of inconsistencies in the XMLs. To help catch structural and value issues more systematically, I decided to formalize the schemas.

Due to the flexibility of the XML files, I found it difficult to express accurate constraints using XSD 1.0. As a result, I upgraded all XSD files to XSD 1.1, which provides more robust constraint capabilities—but please note that this version isn't supported by all XML validators.

With the help of #5184, I corrected some problematic values, but I’ve deliberately avoided making changes to the XML files themselves in this PR.

I'm new to contributing here, so any guidance you can offer—whether it's about testing, workflow, or whether this approach aligns with project goals—is very welcome. If this work is out of scope or not desired, feel free to point me in the right direction or let me know if I should hold off.

Thanks!


Chummer/data/metamagic.xsd
```diff
@@ -1,58 +1,52 @@
...
+                  <xs:all>
+                    <xs:group ref="baseGroup" />
+
+                    <!-- required -->
+                    <xs:element name="adept" type="boolean" />
+                    <xs:element name="magician" type="boolean" />
+
+                    <!-- optional -->
+                    <xs:element ref="bonus" minOccurs="0" />
+                    <xs:element name="required" minOccurs="0" />
```
chummer5a, line 25: This should be a ref to refer to conditions.xsd, shouldn't it?

BM123499: You are corrected, I fixed it on tradition.xsd as well. Unfortunately, xerces assume any type when you don't specify, and it might be overseed easily.

BM123499:
At this point, I believe I'm finished with the XSD updates. I'll be turning my attention back to the importer and the other parts of the project I'm working on. If there’s any feedback or suggestions regarding the schema changes, I’m happy to make adjustments as needed. Thanks again for your time and support!

PR #5210, opened 2025-08-21
Adding FA Magical Compounds and buyable alchemy preparations

Teksura:
Adds the Forbidden Arcana Magical Compounds and "New Alchemical Preparations" from page 194 and 198 respectively. 

Price and Availability is baed off of the following entries:

Screenshot 2025-08-21 at 11 50 59 AM
Screenshot 2025-08-21 at 11 51 10 AM

### DIFF
--- a/Chummer/data/gear.xml
+++ b/Chummer/data/gear.xml
@@ -13544,7 +13544,7 @@
       <id>08b3bf17-e52a-41d4-b023-af6ee5354fe6</id>
       <name>AgHexHex</name>
       <category>Magical Compounds</category>
-      <rating>20</rating>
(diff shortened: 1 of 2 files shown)


DelnarErsike:
The `rating` entries are all set to zero. Is that on purpose? I'd imagine the way we'd want to set it up is to have the Rating be Force + Current Potency. So this'd mean having a minimum rating of 2, an unbounded maximum rating (something like 1000000), and a `ratinglabel` of "Force + Current Potency".

Teksura:
Oh I see how that works now. Requested changes have been applied.

PR #5278, opened 2026-08-07
Security: stored HTML/script injection in character sheets (notes, description, etc.)

skachm:
Character free-text fields (notes, game notes, concept, description, background) were
written into the generated character sheet with `disable-output-escaping="yes"` (in
`Chummer/sheets/xt.PreserveHtml.xslt`). Sheets are rendered in the MSHTML `WebBrowser`
control from a local file and are also piped to wkhtmltopdf for PDF export — both of
which execute HTML/JavaScript. As a result, a shared `.chum5` save or custom-data XML
could run arbitrary script on a recipient's machine when they preview, print, or export
the character.

Changes:
- Remove `disable-output-escaping` from `xt.PreserveHtml.xslt` so those fields are
  HTML-escaped. Line breaks are still preserved (notes are plain multiline text; each
  newline still emits a `<br/>`). Arbitrary HTML formatting in notes is intentionally no
  longer supported.
- Defense in depth for the PDF path: pass
  `--disable-javascript --disable-local-file-access --disable-external-links` to
  wkhtmltopdf in `PdfConvert.cs`, so untrusted sheet HTML can't run script or read
  local/remote resources during PDF generation.

### DIFF
--- a/Chummer/PdfConvert.cs
+++ b/Chummer/PdfConvert.cs
@@ -243,6 +243,8 @@ namespace Codaxy.WkHtmlToPdf
                                                           out StringBuilder sbdParams))
             {
                 sbdParams.Append("--page-size A4 ", "--disable-smart-shrinking ");
+                // SECURITY: sheet HTML is untrusted (character notes/description/etc.); block scripting, local-file reads, and remote links.
+                sbdParams.Append("--disable-javascript ", "--disable-local-file-access ", "--disable-external-links ");
 
                 if (!string.IsNullOrEmpty(document.HeaderUrl))
                 {
--- a/Chummer/sheets/xt.PreserveHtml.xslt
+++ b/Chummer/sheets/xt.PreserveHtml.xslt
@@ -1,12 +1,12 @@
 <?xml version="1.0" encoding="UTF-8" ?>
-<!-- Preserve Line Breaks Template that also preserves Html tag data -->
+<!-- Preserve line breaks (newlines -> <br/>). SECURITY: keep escaping ON; do not add disable-output-escaping (output is rendered by MSHTML/wkhtmltopdf). -->
 <!-- Version -500 -->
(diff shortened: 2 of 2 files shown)


Chummer/sheets/xt.PreserveHtml.xslt
```diff
@@ -15,7 +15,7 @@
         </xsl:call-template>
       </xsl:when>
       <xsl:otherwise>
-        <xsl:value-of select="$text" disable-output-escaping="yes" />
+        <xsl:value-of select="$text" />
```
DelnarErsike, line 18: Did you check to make sure that all sheets all remain functional afterwards? There's a lot of weird transformations of RTF to HTML that need to happen for things like character background and biography sections where line breaks do indeed need be un-escaped to preserve formatting in the sheet.

Chummer/PdfConvert.cs
```diff
@@ -243,6 +243,8 @@ private static async Task ConvertHtmlToPdfCoreAsync(bool blnSync, PdfDocument do
                                                           out StringBuilder sbdParams))
             {
                 sbdParams.Append("--page-size A4 ", "--disable-smart-shrinking ");
+                // SECURITY: sheet HTML is untrusted (character notes/description/etc.); block scripting, local-file reads, and remote links.
+                sbdParams.Append("--disable-javascript ", "--disable-local-file-access ", "--disable-external-links ");
```
DelnarErsike, line 247: This change will need to be rejected. The Linux printing fix works via local file access, and the ability for users to hide and show specific parts of their character sheet are handled through Javascript. Both of those are too important relative to this "defense in depth" setup.

DelnarErsike (changes requested):
Did you check to make sure that sheets all remain functional after removing `disable-output-escaping`?

